PR #1, opened 2017-01-25
Added PELU Activation Function

joggino:


farizrahman4u:
Can you please add a unit test too? You can use advanced activations tests in Keras as example.

the-moliver:
Also the documentation is incorrect. Please make sure it accurately reflects what the function actually does as well as all the relevant parameters.

farizrahman4u:
LGTM. @the-moliver Can you confirm?

the-moliver:
I believe implementation is wrong. The `neg` part will be non-zero when x>0, which is not how it should work.

farizrahman4u:
I am not entirely familiar with the concept.. but I guess the issue is here ?
```python
return pos + neg
```

Maybe this will fix it ?
```python
return K.switch(x < 0., neg, pos)
```

farizrahman4u:
I am not entirely familiar with the concept.. but I guess the issue is here ?
```python
return pos + neg
```

Maybe this will fix it ?
```python
return K.switch(x < 0., neg, pos)
```

the-moliver:
That would work. And the `K.relu(x)` in the `pos` part could then be replaced by just `x`.

joggino:
Would you be able to alter:
```
else:
    pos = K.relu(x) * (self.alphas / self.betas)
``` 

to:
```
else:
     pos = x * (self.alphas / self.betas)
``` 

Or should I open another pull request?

farizrahman4u:
Done.

PR #2, opened 2017-01-25
Markdown page changes

patyork:
- GUIDELINES.md:
  - A bit added (no merging of your own PRs)
- CONTRIBUTING.md:
  - Made similar to Keras'; for anyone that wants to make a PR
  - Made a note of the license, and that code may be moved from KC to K (or vice versa) with correct attribution.
- README.md:
  - Changed link for contributors to go to CONTRIBUTING.md (makes GUIDELINES only for maintainers)
- .gitignore:
  - ignore `.idea` files (PyCharm python IDE files)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,62 @@
+# On Github Issues and Pull Requests
+
+Found a bug? Want to contribute changes to the codebase? Make sure to read this first.
+
+## Update Your Environment
+
+To easily update Keras: `pip install git+https://www.github.com/fchollet/keras.git --upgrade`
+
+To easily update Keras-Contrib: `pip install git+https://www.github.com/farizrahman4u/keras-contrib.git --upgrade`
+
+To easily update Theano: `pip install git+git://github.com/Theano/Theano.git --upgrade`
+
+To update TensorFlow: See [TensorFlow Installation instructions](https://github.com/tensorflow/tensorflow#installation)
+
+## Bug reporting
+
+Your code doesn't work, **and you have determined that the issue lies with Keras-Contrib**? Follow these steps to report a bug.
+
+1. Your bug may already be fixed. Make sure to update to the current Keras master branch and Keras-Contrib master branch, as well as the latest Theano/TensorFlow master branch.
+
+2. [Search for similar issues](https://github.com/farizrahman4u/keras-contrib/issues?utf8=%E2%9C%93&q=is%3Aissue). Make sure to delete `is:open` on the issue search to find solved tickets as well. It's possible somebody has encountered this bug already. Still having a problem? Open an issue on Github to let us know.
+
+3. Make sure you provide us with useful information about your configuration: what OS are you using? What Keras backend are you using? Are you running on GPU? If so, what is your version of Cuda, of cuDNN? What is your GPU?
+
+4. Provide us with a script to reproduce the issue. This script should be runnable as-is and should not require external data download (use randomly generated data if you need to run a model on some test data). We recommend that you use Github Gists to post your code. Any issue that cannot be reproduced is likely to be closed.
+
+5. If possible, take a stab at fixing the bug yourself --if you can!
+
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Pull Requests
+
+We love pull requests. Here's a quick guide:
+
+1. If your PR introduces a change in functionality, make sure you start by opening an issue to discuss whether the change should be made, and how to handle it. This will save you from having your PR closed down the road! Of course, if your PR is a simple bug fix, you don't need to do that.
+
+2. Ensure that your environment (Keras, Keras-Contrib, and your backend) are up to date. See "Update Your Environment". Create a new branch for your changes.
+
+3. Write the code. This is the hard part! If you are adding a layer, advanced activation, or any other feature which has configurable parameters, please ensure that the feature is searializeable (to allow for saving and loading). For detail on this aspect, please see the Keras ["Writing Your Own Layer"](https://keras.io/layers/writing-your-own-keras-layers/) guide and the source code for the relevant feature type from both Keras and Keras-Contrib.
```
lukedeo: "details"

CONTRIBUTING.md
```diff
@@ -0,0 +1,68 @@
+# On Github Issues and Pull Requests
```
lukedeo: Do we want to make an explicit note here about implementation help? A good percentage of Keras issues seem to be requests for help.

CONTRIBUTING.md
```diff
@@ -0,0 +1,68 @@
...
+
+10. Update the documentation. If introducing new functionality, make sure you include code snippets demonstrating the usage of your new feature.
+
+11. Submit your PR. If your changes have been approved in a previous discussion, and if you have complete (and passing) unit tests, your PR is likely to be merged promptly. Otherwise, well...
+
+## A Note for Contributors
+
+Both Keras-Contrib and Keras operate under the [MIT License](LICENSE). At the discretion of the maintainers of both repositories, code may be moved from Keras-Contrib to Keras and vice versa.
+
+The maintainers will ensure that the proper chain of commits will flow in both directions, with proper attribution of code. Maintainers will also do their best to notify contributors when their work is moved between repositories.
```
lukedeo: Do we have any tooling built up yet around git attribution when things are moved to Keras proper?

CONTRIBUTING.md
```diff
@@ -0,0 +1,68 @@
+# On Github Issues and Pull Requests
+
+Found a bug? Want to contribute changes to the codebase? Make sure to read this first.
```
lukedeo: What should the distinction be for an example being merged here vs. keras proper? Obviously, using functionality from contrib should necessitate it's presence in contrib, but how about underdeveloped or overly niche examples?

CONTRIBUTING.md
patyork: I was specifically avoiding that aspect for now.

CONTRIBUTING.md
patyork: That's kind of TBD. At this point, very little new functionality is being added to Keras core. So, if it is new functionality, then it can probably be put here.

There will still need to be some case-by-case weighing on the maintainers' part to determine if something is worth adding. Step 1 of the Pull Requests section reminds people to open an issue first, to see if it is worth implementing.

CONTRIBUTING.md
patyork: I've done a proof of concept, and @fchollet makes a good point that git lets us do anything we want (pretty much).

It's basically just a process of a) finding the relevant commits, which could be the hard part, b) paring the commits down to remove code not relevant to whatever is moving to Keras proper, c) saving these commits as a patch, and them applying the patch to Keras d) optional, make additional changes yourself as a new commit to clean anything up.

farizrahman4u:
@patyork Great stuff! Thanks.

CONTRIBUTING.md
lukedeo: Yeah I really would love to avoid all requests for help here, not sure what the best way to do that is. A big flashing GIF might be a bit too off-putting, hah

PR #3, opened 2017-01-26
Update CONTRIBUTING.md

farizrahman4u:
When 'q' param is given in url, github doesn't add the 'is:open' argument on its own.


kemaswill:
It's OK to delete that statement, but I think it's better to remind the users to check the closed issues as well.

Issue #4, opened 2017-01-26
Refine the README.md


kemaswill:
I think we'd better demonstrate the purpose of this new keras-contrib repository in the README.md. From my point of view, the purpose is to:

1. provide an official keras extension repository for the keras community, which contains newest layers, activations, loss functions, optimizers, and so on.
2. keep the original keras repository clean, only the most valuable functions will be finally merged into the original keras repository.

Any suggestions?

farizrahman4u:
@patyork

PR #5, opened 2017-01-26
Readme updates

patyork:
Added a bit more on "why a contrib repo" and an example script (now that there is an actual contribution!).

Maintainers can edit directly if there is a typo.

Ref #4


the-moliver:
LGTM

PR #11, opened 2017-02-07
Add extract_patches as a new backend op

Dref360:


tests/keras_contrib/backend/test_backends.py
```diff
@@ -72,5 +72,21 @@ def check_composed_tensor_operations(first_function_name, first_function_args,
...
+    def test_extract(self):
+        for input_shape in [(1, 3, 40, 40), (1, 3, 10, 10)]:
+            for kernel_shape in [2, 5]:
+                xval = np.random.random(input_shape)
+
+                kernel = [kernel_shape, kernel_shape]
+                strides = [kernel_shape, kernel_shape]
+                xth = KTH.variable(xval)
+                xtf = KTF.variable(xval)
+                ztf = KTF.eval(KTF.extract_image_patches(xtf, kernel, strides, dim_ordering='th', border_mode="valid"))
```
the-moliver: Can you add a test for `dim_ordering='tf'` as well. Otherwise LGTM

tests/keras_contrib/backend/test_backends.py
```diff
@@ -72,5 +72,21 @@ def check_composed_tensor_operations(first_function_name, first_function_args,
...
+                zth = KTH.eval(KTH.extract_image_patches(xth, kernel, strides, dim_ordering='th', border_mode="valid"))
```
the-moliver: Tests of the other border_modes would be good as well.

the-moliver:
Looks like you just need to PEP8-ify it now

PR #12, opened 2017-02-07
Revert to last correct implementation

the-moliver:
Fixing implementation by changing `K.relu(-x)` to `-K.relu(-x)` caused tests to fail for some reason. Reverting to last correct implementation that doesn't fail tests


patyork:
So, I haven't checked build to be 100% sure, but up until this morning there was a bug in Theano causing all Travis builds to fail everywhere (that pull Theano from git), including the nightly job/PR jobs here and at Keras itself.

Redo it as a PR and let the Travis build run again. Unless that negative sign actually broke something, the build should pass.

PR #14, opened 2017-02-07
Deconvolution3D

gvtulder:
Someone suggested I submit this PR fchollet/keras#4457 to the keras-contrib repository. Here it is.

It's basically a copy of `Deconvolution2D`, but then for 3D.


the-moliver:
Nice job!

PR #16, opened 2017-02-08
Add the example of Wasserstein GAN

bobchennan:
The major code is consistent with the ACGAN example for mnist.

I mainly change it to the Wasserstein GAN and add a progbar indicator to show the current loss value.

### DIFF
--- a/examples/mnist_wgan.py
+++ b/examples/mnist_wgan.py
@@ -1,0 +1,336 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Train an Wasserstein Generative Adversarial Network (WGAN)
+on the MNIST dataset.
+See https://arxiv.org/abs/1610.09585 for more details about ACGAN.
+See https://arxiv.org/abs/1701.07875 for more details about WACGAN.
+
+You should start to see reasonable images after ~3 epochs.
+According to the paper, the performance is highly related
+to the discriminator loss.
+You should use a GPU, as the convolution-heavy operations
+are very slow on the CPU.
+Prefer the TensorFlow backend if you plan on iterating,
+as the compilation time can be a blocker using Theano.
+
+Timings:
+
+Hardware           | Backend | Time / Epoch
+-------------------------------------------
+ CPU               | TF      | 3 hrs
+ Titan X (maxwell) | TF      | 4 min
+ Titan X (maxwell) | TH      | 7 min
+
+Consult https://github.com/bobchennan/Wasserstein-GAN-Keras for
+more information and example output
+The original ACGAN implementation can be found in
+https://github.com/lukedeo/keras-acgan
+More tricks to train GAN can be found in https://github.com/soumith/ganhacks
+"""
+from __future__ import print_function
+
+from collections import defaultdict
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
+from PIL import Image
+
+from six.moves import range
+
+import keras.backend as K
+from keras.datasets import mnist
+from keras.layers import Input, Dense, Reshape, Flatten
+from keras.layers import Dropout, Embedding, merge
+from keras.layers.advanced_activations import LeakyReLU
+from keras.layers.convolutional import UpSampling2D, Convolution2D
+from keras.layers.noise import GaussianNoise
+from keras.models import Sequential, Model
+from keras.optimizers import Adam, SGD
+from keras.backend.common import _EPSILON
+from keras.utils.generic_utils import Progbar
+import numpy as np
+
+np.random.seed(1331)
+
+K.set_image_dim_ordering('th')
+
+
+def modified_binary_crossentropy(target, output):
+    return K.mean(target * output)
+
+
+def build_generator(latent_size):
+    # we will map a pair of (z, L), where z is a latent vector and L is a
+    # label drawn from P_c, to image space (..., 1, 28, 28)
+    cnn = Sequential()
+
+    cnn.add(Dense(1024, input_dim=latent_size))
+    cnn.add(LeakyReLU())
+    cnn.add(Dense(128 * 7 * 7))
+    cnn.add(LeakyReLU())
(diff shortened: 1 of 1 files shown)


examples/mnist_wgan.py
```diff
@@ -0,0 +1,336 @@
...
+from keras.backend.common import _EPSILON
+from keras.utils.generic_utils import Progbar
+import numpy as np
+
+np.random.seed(1331)
+
+K.set_image_dim_ordering('th')
+
+
+def modified_binary_crossentropy(target, output):
```
the-moliver: I think you should just call this `wasserstein_loss` since there is really no relationship to binary cross-entropy any longer.

examples/mnist_wgan.py
```diff
@@ -0,0 +1,336 @@
...
+    latent_size = 100
+
+    # Adam parameters suggested in https://arxiv.org/abs/1511.06434
+    adam_lr = 0.0002
+    adam_beta_1 = 0.5
+
+    # build the discriminator
+    discriminator = build_discriminator()
+    discriminator.compile(
+        optimizer=SGD(clipvalue=0.01),
```
the-moliver: Why are you using 3 different optimizers?

examples/mnist_wgan.py
```diff
@@ -0,0 +1,336 @@
...
+        loss=[modified_binary_crossentropy, 'sparse_categorical_crossentropy']
+    )
+
+    # build the generator
+    generator = build_generator(latent_size)
+    generator.compile(optimizer=Adam(lr=adam_lr, beta_1=adam_beta_1),
```
the-moliver: Why are you using 3 different optimizers?

examples/mnist_wgan.py
```diff
@@ -0,0 +1,336 @@
...
+
+    # get a fake image
+    fake = generator([latent, image_class])
+
+    # we only want to be able to train generation for the combined model
+    discriminator.trainable = False
+    fake, aux = discriminator(fake)
+    combined = Model(input=[latent, image_class], output=[fake, aux])
+
+    combined.compile(
```
the-moliver: Why are you using 3 different optimizers?

examples/mnist_wgan.py
```diff
@@ -0,0 +1,336 @@
...
+
+    X_test = (X_test.astype(np.float32) - 127.5) / 127.5
+    X_test = np.expand_dims(X_test, axis=1)
+
+    nb_train, nb_test = X_train.shape[0], X_test.shape[0]
+
+    train_history = defaultdict(list)
+    test_history = defaultdict(list)
+
+    for epoch in range(nb_epochs):
```
the-moliver: You don't appear to be using the training schedule recommended from the paper. Maybe it would be a good idea for a canonical example like this unless there's a good reason not to.

examples/mnist_wgan.py
bobchennan: I will try to use the same optimizer. Since for the origin GAN it is recommended to use weaker optimizer for discriminator, I didn't change it.

examples/mnist_wgan.py
bobchennan: For discriminator we need to limit the gradient or the absolute value so it makes difference. I think two optimizers are needed.

examples/mnist_wgan.py
bobchennan: Could you please give me more explanation about the difference?

examples/mnist_wgan.py
the-moliver: In the WGAN paper, they recommend using RMSprop for everything, and clipping the weight values of the discriminator. You are using gradient clipping which is different. You should use a constraint instead.

examples/mnist_wgan.py
the-moliver: Check out the training schedule used here: https://github.com/martinarjovsky/WassersteinGAN/blob/master/main.py
Basically, you want to train the discriminator more initially, and then train it a lot every so often so that it maintains good performance.

examples/mnist_wgan.py
bobchennan: Yes but this implementation in Keras is a little bit tricky since there are no such constraint. 
One thing we can do is get all parameters and clip them.

Instead I think if you limited the gradients which is corresponding to the updates, you can also limit the range of the parameters for discriminator.

the-moliver: No it doesn't work like that. Even if you clip the gradients the parameters can grow without bound given enough updates. You can make a custom constraint if needed. Actually I'd recommend creating this constraint as a new PR and then using it here.

examples/mnist_wgan.py
bobchennan: I see. I read that before but since MNIST is a small dataset I think 1 iteration may be enough for discriminator.

I will try to use more iterations.

examples/mnist_wgan.py
bobchennan: Yeah I will try to submit a new PR for keras. But for this constrain is there any other cases that people will use it?

the-moliver: You can submit here instead for the time being

the-moliver: I created a new clip constraint in keras-contrib/constraints.py you can use.

tboquet:
@bobchennan is this PR still an active project?

bobchennan:
Sorry but recently I am attending a conference. The result on MNIST is not outstanding and the training process is slow (for example you need to train discriminator 100 times in the first several iterations). I am still looking for better implementation.

tboquet:
Ok, if I have time I'll try to work on this. We can wait for better results, do you know when you will be able to improve the PR?

MarkSeygan:
I thought that the discriminator weights should be clamped. Maybe I am mistaken but gradient clipping doesn't constraint weights. Thanks for enlightening :)

the-moliver:
I'm going to close this for now in favor of the Improved WGAN example.

PR #21, opened 2017-02-13
Fix deconv3d in TF backend (correct kernel transpose)

gvtulder:
The `deconv3d` function in the TensorFlow backend reordered the kernel dimensions in the wrong way. It was supposed to swap the input and output channels, but instead it was mixing the channel dimensions and spatial dimensions.

There's now a better test (for `deconv3d`, I didn't improve the original `deconv2d` test) that detects this problem by using a different filter size in each dimension.

See also @gattia's comments on fchollet/keras#4457.

This is a TensorFlow-specific problem, the Theano `deconv3d` works and worked fine.


the-moliver:
Thanks! LGTM!

PR #24, opened 2017-02-15
Add Wide Residual Networks

titu1994:
Add Wide Residual Networks again due to a merge mess as well as an example script. 

Also makes a change to the DenseNet docs about its score.


kemaswill:
LGTM.

PR #25, opened 2017-02-16
Added a Linear Chain CRF layer and a text chunking example (parellel PR)

phipleg:
Adds Linear Chain Conditional Random Fields, the data set CoNLL 2000, and a text chunking example.

This PR is the parallel to [https://github.com/fchollet/keras/pull/4621](https://github.com/fchollet/keras/pull/4621). The discussion startet originally in issue [https://github.com/fchollet/keras/issues/4090](https://github.com/fchollet/keras/issues/4090). 

Contributions and comments so far from:
* @linxihui
* @fchollet 
* @nreimers
* @tboquet
* @napsternxg

I would be very happy to receive more feedback! :smile:


xyz11:
Does it support class_weight/sample_weight right now? (seems not). Will the support be added in the future?

PR #26, opened 2017-02-19
add dssim objective

Dref360:


tests/keras_contrib/test_objectives.py
```diff
@@ -25,5 +24,32 @@ def test_objective_shapes_2d():
         assert K.eval(objective_output).shape == (6,)
 
 
+def test_dssim_same():
+    x = np.random.random_sample(30 * 30 * 3).reshape([1, 30, 30, 3])
+    x1 = KC.variable(x)
+    loss = objectives.DSSIMObjective(1)
+    np.allclose([0.0], KC.eval(loss(x1, x1)))
```
patyork: `np.allclose` won't fail the test if it returns False - you could use the numpy function `from numpy.testing import assert_allclose` or just apply an asser(... == True)

tests/keras_contrib/test_objectives.py
```diff
@@ -25,5 +24,32 @@ def test_objective_shapes_2d():
...
+
+
+def test_dssim_opposite():
+    x = np.zeros([1, 30, 30, 3])
+    x1 = KC.variable(x)
+    y = np.ones([1, 30, 30, 3])
+    y1 = KC.variable(y)
+    loss = objectives.DSSIMObjective(1)
+    np.allclose([0.5], KC.eval(loss(x1, y1)))
```
patyork: `np.allclose` won't fail the test if it returns False - you could use the numpy function `from numpy.testing import assert_allclose` or just apply an asser(... == True)

keras_contrib/objectives.py
```diff
@@ -1,10 +1,62 @@
...
+        Difference of Structural Similarity (DSSIM loss function). Clipped between 0 and 0.5
+        Note : You shoul add a regularization term like a l2 loss in addition to this one.
+        :param batch_size: Batch size used in the model
+        :param k1: Parameter of the SSIM (default 0.01)
+        :param k2: Parameter of the SSIM (default 0.03)
+        :param kernel_size: Size of the sliding window (default 3)
+        :param max_value: Max value of the output (default 1.0)
+        """
+        self.__name__ = "DSSIMObjective"
+        self.batch_size = batch_size
```
patyork: Does this handle the case where the final batch in an epoch is < batch_size (e.g. num_samples % batch_size != 0)?

keras_contrib/objectives.py
```diff
@@ -1,10 +1,60 @@
...
+        bs, w, h, c1, c2, c3 = self.__int_shape(patches_pred)
+        patches_pred = KC.reshape(patches_pred, [-1, w, h, c1 * c2 * c3])
+        patches_true = KC.reshape(patches_true, [-1, w, h, c1 * c2 * c3])
+        u_true = KC.mean(patches_true, axis=-1)
+        u_pred = KC.mean(patches_pred, axis=-1)
+        var_true = K.var(patches_true, axis=-1)
+        var_pred = K.var(patches_pred, axis=-1)
+        std_true = K.sqrt(var_true + KC.epsilon())
+        std_pred = K.sqrt(var_pred + KC.epsilon())
+        c1 = (self.k1 * self.max_value) ** 2
```
the-moliver: Use `K.square` instead of `** 2`

Issue #17, opened 2017-02-12
Call for Contribution


kemaswill:
Contributions w.r.t. the following layers and examples are welcomed:

- Layers
   - [Layer Normalization](https://arxiv.org/abs/1607.06450)
   - [PELU](https://arxiv.org/abs/1605.09332) [DONE!]
   - [Batch Renormalization](https://arxiv.org/abs/1702.03275)
- Examples
  - [DenseNets](https://arxiv.org/abs/1608.06993v2) [DONE!]
  - [Residual of Residual Networks](https://arxiv.org/pdf/1608.02908v1.pdf) [DONE!]
  - [Wide ResNets](https://arxiv.org/abs/1605.07146v1)[DONE!]
  - [ResNet in ResNet](https://arxiv.org/abs/1603.08029)

titu1994:
I'd be happy to add PRs, but I need to know what format would suite this
repo.

Generally applications are one offs, they build a specific model, load
weights and then provide the model for the end user.

The models I have are builder types, meaning there are multiple parameter
which must be supplied.

On top of that, the models are not trained on ImageNet. Only Cifar 10 or
100. Also, these weights are for the mid level models, since I don't have a
big enough GPU to train the largest models in the papers.

Any suggestions on these issues ?

Also, while we are discussing this, what about the callback builder for
Snapshot Ensemble? It is achieved via a callback list and not a single
callback.

Should I keep the builder class to simplify the usage, or just put the
callbacks separately and the schedule separately and the user must take the
trouble to set up the model ? Personally I prefer the builder approach.

Somshubra Majumdar

kemaswill:
> The models I have are builder types, meaning there are multiple parameter
which must be supplied.

I think you can provide a builder types, defining several functions with different parameters, such as  def wrn_28_10(): return wrn.create_wide_residual_network(ip, nb_classes=10, N=4, k=10, dropout=0.0, verbose=1) and so on

> On top of that, the models are not trained on ImageNet. Only Cifar 10 or
100.

I think that's fine.

> Also, while we are discussing this, what about the callback builder for
Snapshot Ensemble?

Let's discuss this on the [original issue](https://github.com/farizrahman4u/keras-contrib/issues/10)

titu1994:
@kemaswill Thanks for the clarification. 

Should I add an example script of how to use the DenseNet model as well? As in a script which shows how the DenseNet-40-12 model was trained on the CIFAR-10 dataset?

[...]

PR #28, opened 2017-02-19
Added Batch Renormalization Layer

titu1994:
Adds Batch Renormalization layer from the paper [Batch Renormalization: Towards Reducing Minibatch Dependence in Batch-Normalized Models](https://arxiv.org/pdf/1702.03275.pdf) by Sergey Ioffe

### DIFF
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -1,7 +1,8 @@
 from keras.engine import Layer, InputSpec
 from .. import initializations, regularizers
 from .. import backend as K
-from keras.layers import BatchNormalization
+
+import numpy as np
 
 
 class BatchRenormalization(Layer):
@@ -66,8 +67,9 @@ class BatchRenormalization(Layer):
     """
 
     def __init__(self, epsilon=1e-3, mode=0, axis=-1, momentum=0.99,
-                 weights=None, beta_init='zero', gamma_init='one',
-                 gamma_regularizer=None, beta_regularizer=None, **kwargs):
+                 r_max_val=3., d_max_val=5., weights=None, beta_init='zero',
+                 gamma_init='one', gamma_regularizer=None, beta_regularizer=None,
+                 **kwargs):
         self.supports_masking = True
         self.beta_init = initializations.get(beta_init)
         self.gamma_init = initializations.get(gamma_init)
@@ -78,8 +80,8 @@ class BatchRenormalization(Layer):
         self.gamma_regularizer = regularizers.get(gamma_regularizer)
         self.beta_regularizer = regularizers.get(beta_regularizer)
         self.initial_weights = weights
-        self.r_max = 3.
-        self.d_max = 5.
+        self.r_max_value = r_max_val
+        self.d_max_value = d_max_val
         if self.mode == 0:
             self.uses_learning_phase = True
         super(BatchRenormalization, self).__init__(**kwargs)
@@ -103,6 +105,19 @@ class BatchRenormalization(Layer):
                                            name='{}_running_std'.format(self.name),
                                            trainable=False)
 
+        # self.r_max = self.add_weight((1,), initializer='one',
+        #                              name='{}_r_max'.format(self.name),
+        #                              trainable=False)
+        self.r_max = K.variable(np.ones((1,)), name='{}_r_max'.format(self.name))
+
+        # self.d_max = self.add_weight((1,), initializer='zero',
+        #                              name='{}_d_max'.format(self.name),
+        #                              trainable=False)
+
+        self.d_max = K.variable(np.zeros((1,)), name='{}_d_max'.format(self.name))
+
+        self.t = K.variable(np.zeros((1,)), name='{}_t'.format(self.name))
+
         if self.initial_weights is not None:
             self.set_weights(self.initial_weights)
             del self.initial_weights
@@ -125,11 +140,13 @@ class BatchRenormalization(Layer):
                 mean_batch = K.mean(x, axis=reduction_axes)
                 std_batch = K.sqrt(K.var(x, axis=reduction_axes) + self.epsilon)
 
+            r_max_val = K.get_value(self.r_max)
             r = std_batch / self.running_std
-            r = K.stop_gradient(K.clip(r, 1 / self.r_max, self.r_max))
+            r = K.stop_gradient(K.clip(r, 1 / r_max_val, r_max_val))  # clip from [1 / r_max, 1] to avoid gradient explosion
 
+            d_max_val = K.get_value(self.d_max)
             d = (mean_batch - self.running_mean) / self.running_mean
-            d = K.stop_gradient(K.clip(d, -self.d_max, self.d_max))
+            d = K.stop_gradient(K.clip(d, -d_max_val, d_max_val))
 
             if sorted(reduction_axes) == range(K.ndim(x))[:-1]:
                 # now compute the re-normalized batch norm
@@ -153,6 +170,16 @@ class BatchRenormalization(Layer):
             # explicit update to moving mean and standard deviation
             self.add_update([K.moving_average_update(self.running_mean, mean_batch, self.momentum),
                              K.moving_average_update(self.running_std, std_batch, self.momentum)], x)
+
+            # # update r_max and d_max
+            t_val = K.get_value(self.t)
+            r_val = self.r_max_value / (1 + (self.r_max_value - 1) * np.exp(-t_val))
+            d_val = self.d_max_value / (1 + ((self.d_max_value / 1e-3) - 1) * np.exp(-(2 * t_val)))
+            t_val += 1.
+
+            self.add_update([K.update(self.r_max, r_val),
+                             K.update(self.d_max, d_val),
+                             K.update(self.t, t_val)], x)
 
             if self.mode == 0:
                 if sorted(reduction_axes) == range(K.ndim(x))[:-1]:
@@ -190,7 +217,6 @@ class BatchRenormalization(Layer):
             # x_normed = ((x_normed * r) + d) * self.gamma + self.beta
             raise ValueError('Batch Renormalization does not support mode=1')
 
-
         return x_normed
 
     def get_config(self):
@@ -201,6 +227,7 @@ class BatchRenormalization(Layer):
                   'beta_regularizer': self.beta_regularizer.get_config() if self.beta_regularizer else None,
                   'momentum': self.momentum,
                   'r_max': self.r_max,
-                  'd_max': self.d_max}
+                  'd_max': self.d_max,
+                  't': self.t}
         base_config = super(BatchRenormalization, self).get_config()
         return dict(list(base_config.items()) + list(config.items()))
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -39,6 +39,7 @@ class BatchRenormalization(Layer):
             of the data, for feature-wise normalization.
         r_max_value: Upper limit of the value of r_max.
         d_max_value: Upper limit of the value of d_max.
+        t_scale: Parameter to scale time steps.
         weights: Initialization weights.
             List of 2 Numpy arrays, with shapes:
             `[(input_shape,), (input_shape,)]`
@@ -69,7 +70,7 @@ class BatchRenormalization(Layer):
     """
 
     def __init__(self, epsilon=1e-3, mode=0, axis=-1, momentum=0.99,
-                 r_max_val=3., d_max_val=5., weights=None, beta_init='zero',
+                 r_max_val=3., d_max_val=5., t_scale=0.1, weights=None, beta_init='zero',
                  gamma_init='one', gamma_regularizer=None, beta_regularizer=None,
                  **kwargs):
         self.supports_masking = True
@@ -84,6 +85,7 @@ class BatchRenormalization(Layer):
         self.initial_weights = weights
         self.r_max_value = r_max_val
         self.d_max_value = d_max_val
+        self.t_delta = t_scale
         if self.mode == 0:
             self.uses_learning_phase = True
         super(BatchRenormalization, self).__init__(**kwargs)
@@ -166,7 +168,7 @@ class BatchRenormalization(Layer):
             t_val = K.get_value(self.t)
             r_val = self.r_max_value / (1 + (self.r_max_value - 1) * np.exp(-t_val))
             d_val = self.d_max_value / (1 + ((self.d_max_value / 1e-3) - 1) * np.exp(-(2 * t_val)))
-            t_val += 1.
+            t_val += float(self.t_delta)
 
             self.add_update([K.update(self.r_max, r_val),
                              K.update(self.d_max, d_val),
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -174,6 +174,26 @@ class BatchRenormalization(Layer):
                              K.update(self.d_max, d_val),
                              K.update(self.t, t_val)], x)
 
+            if sorted(reduction_axes) == range(K.ndim(x))[:-1]:
+                x_normed_running = K.batch_normalization(
+                    x, self.running_mean, self.running_std,
+                    self.beta, self.gamma,
+                    epsilon=self.epsilon)
+            else:
+                # need broadcasting
+                broadcast_running_mean = K.reshape(self.running_mean, broadcast_shape)
+                broadcast_running_std = K.reshape(self.running_std, broadcast_shape)
+                broadcast_beta = K.reshape(self.beta, broadcast_shape)
+                broadcast_gamma = K.reshape(self.gamma, broadcast_shape)
+                x_normed_running = K.batch_normalization(
+                    x, broadcast_running_mean, broadcast_running_std,
+                    broadcast_beta, broadcast_gamma,
+                    epsilon=self.epsilon)
+
+            # pick the normalized form of x corresponding to the training phase
+            # for batch renormalization, inference time remains same as batchnorm
+            x_normed = K.in_train_phase(x_normed, x_normed_running)
+
         elif self.mode == 1:
             # sample-wise normalization
             m = K.mean(x, axis=self.axis, keepdims=True)
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -105,14 +105,7 @@ class BatchRenormalization(Layer):
                                            name='{}_running_std'.format(self.name),
                                            trainable=False)
 
-        # self.r_max = self.add_weight((1,), initializer='one',
-        #                              name='{}_r_max'.format(self.name),
-        #                              trainable=False)
         self.r_max = K.variable(np.ones((1,)), name='{}_r_max'.format(self.name))
-
-        # self.d_max = self.add_weight((1,), initializer='zero',
-        #                              name='{}_d_max'.format(self.name),
-        #                              trainable=False)
 
         self.d_max = K.variable(np.zeros((1,)), name='{}_d_max'.format(self.name))
 
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -130,12 +130,8 @@ class BatchRenormalization(Layer):
             broadcast_shape = [1] * len(input_shape)
             broadcast_shape[self.axis] = input_shape[self.axis]
 
-            if K.backend() == 'tf':
-                mean_batch, var = K.KTF.nn.moments(x, reduction_axes, shift=None, name=None, keep_dims=False)
-                std_batch = (K.sqrt(var + self.epsilon))
-            else:
-                mean_batch = K.mean(x, axis=reduction_axes)
-                std_batch = K.sqrt(K.var(x, axis=reduction_axes) + self.epsilon)
+            mean_batch, var_batch = K.moments(x, reduction_axes, shift=None, keep_dims=False)
+            std_batch = (K.sqrt(var_batch + self.epsilon))
 
             r_max_val = K.get_value(self.r_max)
             r = std_batch / (K.sqrt(self.running_std + self.epsilon))
@@ -214,7 +210,7 @@ class BatchRenormalization(Layer):
             t_val = K.get_value(self.t)
             r_val = self.r_max_value / (1 + (self.r_max_value - 1) * np.exp(-t_val))
             d_val = self.d_max_value / (1 + ((self.d_max_value / 1e-3) - 1) * np.exp(-(2 * t_val)))
-            t_val += 1.
+            t_val += float(self.t_delta)
 
             self.add_update([K.update(self.r_max, r_val),
                              K.update(self.d_max, d_val),
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -119,26 +119,26 @@ class BatchRenormalization(Layer):
             broadcast_shape[self.axis] = input_shape[self.axis]
 
             if K.backend() == 'tf':
-                mean, var = K.KTF.nn.moments(x, reduction_axes, shift=None, name=None, keep_dims=False)
-                std = (K.sqrt(var + self.epsilon))
+                mean_batch, var = K.KTF.nn.moments(x, reduction_axes, shift=None, name=None, keep_dims=False)
+                std_batch = (K.sqrt(var + self.epsilon))
             else:
-                mean = K.mean(x, axis=reduction_axes)
-                std = K.sqrt(K.var(x, axis=reduction_axes) + self.epsilon)
+                mean_batch = K.mean(x, axis=reduction_axes)
+                std_batch = K.sqrt(K.var(x, axis=reduction_axes) + self.epsilon)
 
-            r = std / self.running_std
+            r = std_batch / self.running_std
             r = K.stop_gradient(K.clip(r, 1 / self.r_max, self.r_max))
 
-            d = (mean - self.running_mean) / std
+            d = (mean_batch - self.running_mean) / self.running_mean
             d = K.stop_gradient(K.clip(d, -self.d_max, self.d_max))
 
             if sorted(reduction_axes) == range(K.ndim(x))[:-1]:
                 # now compute the re-normalized batch norm
-                x_normed = (x - mean) / std
+                x_normed = (x - mean_batch) / std_batch
                 x_normed = (x_normed * r + d) * self.gamma + self.beta
             else:
                 # need broadcasting
-                broadcast_mean = K.reshape(mean, broadcast_shape)
-                broadcast_std = K.reshape(std, broadcast_shape)
+                broadcast_mean = K.reshape(mean_batch, broadcast_shape)
+                broadcast_std = K.reshape(std_batch, broadcast_shape)
                 broadcast_r = K.reshape(r, broadcast_shape)
                 broadcast_d = K.reshape(d, broadcast_shape)
                 broadcast_beta = K.reshape(self.beta, broadcast_shape)
@@ -151,8 +151,8 @@ class BatchRenormalization(Layer):
                 x_normed = (x_normed * broadcast_r + broadcast_d) * broadcast_gamma + broadcast_beta
 
             if self.mode == 0:
-                self.add_update([K.moving_average_update(self.running_mean, mean, self.momentum),
-                                 K.moving_average_update(self.running_std, std, self.momentum)], x)
+                self.add_update([K.moving_average_update(self.running_mean, mean_batch, self.momentum),
+                                 K.moving_average_update(self.running_std, std_batch, self.momentum)], x)
 
                 if sorted(reduction_axes) == range(K.ndim(x))[:-1]:
                     x_normed_running = K.batch_normalization(
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -39,7 +39,7 @@ class BatchRenormalization(Layer):
             of the data, for feature-wise normalization.
         r_max_value: Upper limit of the value of r_max.
         d_max_value: Upper limit of the value of d_max.
-        t_scale: Parameter to scale time steps.
+        t_delta: At each iteration, increment the value of t by t_delta.
         weights: Initialization weights.
             List of 2 Numpy arrays, with shapes:
             `[(input_shape,), (input_shape,)]`
@@ -70,7 +70,7 @@ class BatchRenormalization(Layer):
     """
 
     def __init__(self, epsilon=1e-3, mode=0, axis=-1, momentum=0.99,
-                 r_max_val=3., d_max_val=5., t_scale=0.1, weights=None, beta_init='zero',
+                 r_max_val=3., d_max_val=5., t_delta=0.1, weights=None, beta_init='zero',
                  gamma_init='one', gamma_regularizer=None, beta_regularizer=None,
                  **kwargs):
         self.supports_masking = True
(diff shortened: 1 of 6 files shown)


keras_contrib/layers/normalization.py
```diff
@@ -1,3 +1,235 @@
...
+        if self.mode == 0 or self.mode == 2:
+            assert self.built, 'Layer must be built before being called'
+            input_shape = K.int_shape(x)
+
+            reduction_axes = list(range(len(input_shape)))
+            del reduction_axes[self.axis]
+            broadcast_shape = [1] * len(input_shape)
+            broadcast_shape[self.axis] = input_shape[self.axis]
+
+            if K.backend() == 'tf':
```
the-moliver: Why can't you just use `K.mean` and `K.std` without back-end checking?

keras_contrib/layers/normalization.py
```diff
@@ -1,3 +1,235 @@
...
+
+                x_normed_batch = (x - broadcast_mean) / broadcast_std
+                x_normed = (x_normed_batch * broadcast_r + broadcast_d) * broadcast_gamma + broadcast_beta
+
+            # explicit update to moving mean and standard deviation
+            self.add_update([K.moving_average_update(self.running_mean, mean_batch, self.momentum),
+                             K.moving_average_update(self.running_std, std_batch ** 2, self.momentum)], x)
+
+            # update r_max and d_max
+            t_val = K.get_value(self.t)
```
the-moliver: This schedule for updating `r` and `d` should be adjustable by the user

keras_contrib/layers/normalization.py
```diff
@@ -1,3 +1,235 @@
...
+            r_val = self.r_max_value / (1 + (self.r_max_value - 1) * np.exp(-t_val))
+            d_val = self.d_max_value / (1 + ((self.d_max_value / 1e-3) - 1) * np.exp(-(2 * t_val)))
+            t_val += 1.
+
+            self.add_update([K.update(self.r_max, r_val),
+                             K.update(self.d_max, d_val),
+                             K.update(self.t, t_val)], x)
+
+            if self.mode == 0:
```
the-moliver: I think there is something wrong with your mode 0. For prediction mode 0 in regular batch norm it uses the running mean and std and thus should be the same as batch renormalization. Mode 2 in regular batch norm is if you wanted to not use the running mean and std in predictions, so I'm not even sure it makes sense in the batch renormalization case.

keras_contrib/layers/normalization.py
titu1994: Because tensorflow has a nice direct method called moments. Theano has no equivalent. So for efficiency purposes on tensorflow, I use its specific call, and on Theano, I have no choice but to compute it manually.

the-moliver: Is there really a significant speed difference? Backend specific calls should only be used where absolutely necessary and I'm not sure this qualifies.

keras_contrib/layers/normalization.py
titu1994: It is. The user can specify r_max_value and d_max_value. That will change the scale. If you mean the rate at which the value changes (the algorithm), I tried out 3-4 different algorithms that change the rate, this one is by far the smoothest and fastest. 

For reference: I have tried linear, squared, general population growth, derivative based approaches and finally biological population growth formula. The last one performs the best. 

The author of the paper simply states "For Batch Renorm, we used rmax = 1, dmax = 0 (i.e. simply batchnorm) for the first 5000 training steps, after which these were gradually relaxed to reach rmax = 3 at 40k steps, and dmax = 5 at 25k steps.". Since we cannot access "iteration count" or "epoch" count inside a layer, I had to resort to using my own time counter to perform such a scaling. 

I do not think there is any "simple" way of allowing the user to adjust the rate, and a callback approach is not my idea of good design.

keras_contrib/layers/normalization.py
titu1994: There is. On tensorflow backend, if I take the theano approach, my WRN-16-4 model takes approximately 148 seconds per epoch on mode 2. With the tensorflow implementation, it takes 142 seconds. A small difference, but I am trying to squeeze every second of performance as possible here.

keras_contrib/layers/normalization.py
titu1994: For prediction in mode 0, regular batchnorm and batch renorm both fall back to `x_normed_running` during prediction phase. In training phase, regular batchnorm uses the normal batch norm output `x_normed`, whereas batch renorm uses the renormalized value of `x_normed`.

In mode 2, regular batchnorm and batch renorm both output their respective version of `x_normed` for both training and prediction.

keras_contrib/layers/normalization.py
the-moliver: Yes, I mean the rate of change should be adjustable independently of the max value. The algorithm is fine otherwise.

titu1994: The rate of change is t. If a user really wanted to change it, they can use a callback to set the value of t as shown by @bobchennan, and that will propagate to the updates.

keras_contrib/layers/normalization.py
the-moliver: The point is that in batch renorm `x_normed` should, if things are working as expected, quickly become nearly identical to `x_normed_running` after d and r reach their final values. So then there would be no distinction between mode 0 and 2.

titu1994:
@the-moliver How do I run the tests locally? When I run the test suit from PyCharm at the root test directory, it gives me the following output:
```
Process finished with exit code 0
Empty test suite.
```

keras_contrib/layers/normalization.py
the-moliver: `t` is not a user adjustable parameter, it appears to be an iteration counter (not an epoch counter as you mentioned earlier). You just need a user adjustable parameter that could be multiplied by `t_val` to control the rate of change. Also, based on the plots you showed earlier it looks like `r` and `d` will reach their max after only a few iterations, which doesn't seem right.

keras_contrib/layers/normalization.py
titu1994: x_normed value is dependent on r and d. It will be similar to batch norm only at the very beginning, where r = 1 and d = 0 (as mentioned in the paper). At the final value, x_normed in batch norm is not the same as x_normed in batch renormed. 

From the paper - ": the fixed (for the given minibatch) r and d correct for the fact that the minibatch statistics differ from the population ones. This allows the above layers to observe the “correct” activations – namely, the ones that would be generated by the inference model." 

This suggests that in mode 0, only at iteration 0, where r=1 and d=0, will the test time output of x_normed be equal to x_normed_running (since they are both simply batch norm outputs).

the-moliver:
Hmm not exactly sure. You could just run the tests manually until someone has a better suggestion.

patyork:
There's instructions in the CONTIBUTING.md file - same process as within Keras itself.

titu1994:
@the-moliver @patyork Ran the tests, they are all passing locally, and they should pass through travis as well.

keras_contrib/layers/normalization.py
the-moliver: No, I think you misunderstand the paper. Look at what `r` and `d` do mathematically, assuming no clipping is taking place (which should be the case near their final value as stated in the paper): They cancel out the batch statistics and replace them with the running statistics.

the-moliver:
Yes, but you need to write new tests for this new layer.

tests/keras_contrib/layers/test_normalization.py
```diff
@@ -9,6 +9,117 @@
...
+    model.add(normalization.BatchRenormalization(mode=0, input_shape=(10, 5, 5), axis=1))
+    model.compile(loss='mse', optimizer='sgd')
+
+    X = np.random.normal(loc=5.0, scale=10.0, size=(20, 10, 5, 5))
+    model.fit(X, X, nb_epoch=1, verbose=0)
+    model.predict(X)
+
+
+@keras_test
+def test_batchnorm_mode_0_convnet():
```
patyork: "re" norm

tests/keras_contrib/layers/test_normalization.py
```diff
@@ -9,6 +9,117 @@
...
+    out = model.predict(X)
+    out -= np.reshape(K.eval(norm_m0.beta), (1, 3, 1, 1))
+    out /= np.reshape(K.eval(norm_m0.gamma), (1, 3, 1, 1))
+
+    assert_allclose(np.mean(out, axis=(0, 2, 3)), 0.0, atol=1e-1)
+    assert_allclose(np.std(out, axis=(0, 2, 3)), 1.0, atol=1e-1)
+
+
+@keras_test
+def test_batchnorm_mode_1():
```
patyork: "re" norm

tests/keras_contrib/layers/test_normalization.py
```diff
@@ -9,6 +9,117 @@
...
+        out = (norm_m1.call(K.variable(inp)) - norm_m1.beta) / norm_m1.gamma
+        assert_allclose(K.eval(K.mean(out)), 0.0, atol=1e-1)
+        if inp.std() > 0.:
+            assert_allclose(K.eval(K.std(out)), 1.0, atol=1e-1)
+        else:
+            assert_allclose(K.eval(K.std(out)), 0.0, atol=1e-1)
+
+
+@keras_test
+def test_shared_batchnorm():
```
patyork: "re" norm

titu1994:
Well that's weird, the output locally is : 
```
Using TensorFlow backend.
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library cublas64_80.dll locally
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library cudnn64_5.dll locally
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library cufft64_80.dll locally
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library nvcuda.dll locally
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library curand64_80.dll locally
============================= test session starts =============================
platform win32 -- Python 3.5.2, pytest-2.9.2, py-1.4.31, pluggy-0.3.1 -- D:\Users\Yue\Anaconda3-5\python.exe
cachedir: ..\..\..\.cache
rootdir: D:\Users\Yue\PycharmProjects\keras-contrib, inifile: pytest.ini
collecting ... collected 5 items

test_normalization.py::test_batchrenorm_mode_0_or_2 PASSED
test_normalization.py::test_batchrenorm_mode_0_or_2_twice PASSED
test_normalization.py::test_batchnorm_mode_0_convnet PASSED
test_normalization.py::test_batchnorm_mode_1 PASSED
test_normalization.py::test_shared_batchnorm PASSED

========================== slowest 10 test durations ==========================
48.58s call     tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_0_or_2
23.89s call     tests/keras_contrib/layers/test_normalization.py::test_batchnorm_mode_0_convnet
3.22s call     tests/keras_contrib/layers/test_normalization.py::test_shared_batchnorm
1.74s call     tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_0_or_2_twice
1.58s call     tests/keras_contrib/layers/test_normalization.py::test_batchnorm_mode_1
0.00s setup    tests/keras_contrib/layers/test_normalization.py::test_batchnorm_mode_0_convnet
0.00s teardown tests/keras_contrib/layers/test_normalization.py::test_shared_batchnorm
0.00s teardown tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_0_or_2_twice
0.00s setup    tests/keras_contrib/layers/test_normalization.py::test_shared_batchnorm
0.00s teardown tests/keras_contrib/layers/test_normalization.py::test_batchnorm_mode_1
========================== 5 passed in 79.02 seconds ==========================

```

tests/keras_contrib/layers/test_normalization.py
patyork: Also, this test is failing (hitting NaN)

tests/keras_contrib/layers/test_normalization.py
titu1994: Aah thank you for that. Will look into the errors.

titu1994:
@patyork I don't get why it's getting nan's in travis. On my local pc after re running the tests in both theano and tensorflow I got the same results:
```
Using TensorFlow backend.
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library cublas64_80.dll locally
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library cudnn64_5.dll locally
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library cufft64_80.dll locally
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library nvcuda.dll locally
I c:\tf_jenkins\home\workspace\release-win\device\gpu\os\windows\tensorflow\stream_executor\dso_loader.cc:135] successfully opened CUDA library curand64_80.dll locally
============================= test session starts =============================
platform win32 -- Python 3.5.2, pytest-3.0.6, py-1.4.32, pluggy-0.4.0 -- D:\Users\Yue\Anaconda3-5\python.exe
cachedir: ..\..\..\.cache
rootdir: D:\Users\Yue\PycharmProjects\keras-contrib, inifile: pytest.ini
collecting ... collected 5 items

test_normalization.py::test_batchrenorm_mode_0_or_2 PASSED
test_normalization.py::test_batchrenorm_mode_0_or_2_twice PASSED
test_normalization.py::test_batchrenorm_mode_0_convnet PASSED
test_normalization.py::test_batchrenorm_mode_1 PASSED
test_normalization.py::test_shared_batchrenorm PASSED

========================== slowest 10 test durations ==========================
103.65s call     tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_0_or_2
49.70s call     tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_0_convnet
4.55s call     tests/keras_contrib/layers/test_normalization.py::test_shared_batchrenorm
3.15s call     tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_0_or_2_twice
1.87s call     tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_1
0.00s teardown tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_0_or_2
0.00s setup    tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_0_convnet
0.00s setup    tests/keras_contrib/layers/test_normalization.py::test_batchrenorm_mode_0_or_2
0.00s teardown tests/keras_contrib/layers/test_normalization.py::test_shared_batchrenorm
0.00s setup    tests/keras_contrib/layers/test_normalization.py::test_shared_batchrenorm
========================= 5 passed in 163.00 seconds ==========================
```

patyork:
I reran Travis, same error.

The test is also failing on my machine, so I'll take a quick peek. (Theano dev, float32, GPU, tf ordering default).

keras_contrib/layers/normalization.py
titu1994: Hmm I guess you are right about it getting to its max value very quickly. The problem is I cant obtain the iteration count from the layer. Or even the epoch count. The paper suggests 40k iterations and 25k iterations for r_max and d_max, but since iteration count is also dependent on batch size, I can't really do much.

While I do wan't to follow the paper to a t, it isnt quite possible yet with Keras working as it is. Also, while it is quite weird that in just 5 iterations it is hitting its max value, this approach seems to have similar results to what the paper describes. Without this increment, it hits nan's due to exploding gradients. With such a sharp increase in its max, it still works. 

How about setting the user adjustable parameter to a default value of 0.1 and multiply it with t_val. That would require roughly 50 iterations instead of 5 to reach its final value. I can't test the effects just yet, since my laptop crashed and my nearly finished training of the BRn WRN-16-4 failed so I have to wait for it to complete again (4-5 hours).

titu1994:
@patyork The output for K.eval(K.mean(out) is coming out to be exactly 0.0, as expected, as at t=0, r = 1 and d = 0, so batch renorm output is equal to batch norm output at that point. I don't understand where nan's are coming from. I am using Python 3.5 as well, with tf 1.0 and Keras master.

keras_contrib/layers/normalization.py
titu1994: Hmm I re-read the paper and I see that you are right. At their max values, the paper suggests that x_normed and x_normed_running become equal. So basically, there is no need to even have a mode=2 anymore, since mode=0 should become mode=2 with enough iterations. Should I drop the mode=0 specific code then?

patyork:
It seems to be failing on mode_1 with `input_2` which is a set of zeros.

Without thinking much about it, that sounds an awful lot like a division by zero.

titu1994:
@patyork Mean and std of 0s will be 0s themselves. I've added epsilon in certain areas where this may happen, so hopefully this test will pass. 

The odd thing is, why are my tests passing if they divide by zero?

patyork:
So I've had some odd results with pytest.

All the test started passing unless I changed the keras.json settings; if I swapped back they failed again, but after multiple runs they started passing. I tried `--cache-clear` option to pytest, but didn't seem to work. Perhaps clearing all the temp dirs manually...

titu1994:
@patyork It's working now. Test's passed. It appears to be the calculation for self.running_mean was 0 (since all 0s) hence the division by 0 occurred. Very rare case where every input is 0, but I will add these changes to the main code as well, just in case something like this happens during real execution.

keras_contrib/layers/normalization.py
titu1994: No, looking at it again see the inference diagram at the end of the algorithm. It states y = gamma * (x - running_mean) / (running_std) + beta.

There is no mention of r and d here. The results of my testing this algorithm also show similarities to what the paper described. Validation accuracies of both BatchNorm and BatchRenorm should be very similar if all other things are kept constant. Batch renorm may give a very small boost to the final result however, as is the case in the paper when using Inception v3 over ImageNet. 

This also corresponds to my own readings comparing the scores of WRN-16-4 with Batchnorm and Batchrenorm. Both are behaving similarly, with batch renorm giving slightly better performance. The inference portion is important here and cannot be removed, since inference in BRn should be similar to BN inference.

titu1994:
I think the algorithm is just about done. For a final check, we should wait for the results from my WRN model, whose output will validate the implementation if the two graphs are similar enough.

keras_contrib/layers/normalization.py
the-moliver: That seems reasonable

the-moliver: If go smaller for the default actually. 50 iterations is still tiny, 500 or 5000 is probably better

keras_contrib/layers/normalization.py
the-moliver: You're missing my point again. Please read my comments from the beginning. And then explain what you think the difference between inference in mode 0 and mode 2 is and should be.

titu1994:
@the-moliver @patyork So below is the training graph for batchnorm vs batch renorm in WRN-16-4 on CIFAR 10 without any dropout or data augmentation for 100 epochs. It overfits obviously, but they both overfit similarly. The training graph is almost exactly the same for both. On the other side, batch renorm gets a slightly highers accuracy score than plain batchnorm.

batchnorm vs renorm mode 0

titu1994:
@the-moliver I ran some tests in which I used different values of t_delta - [1. (initial case), 0.1, 0.01, 1e-3, 1e-4]

What I found was that, eventually, they all landed up with nearly the same scores, but the major difference was how many epochs it took to reach there. 

For example, taking WRN-16-4 model with BRn mode 2 as the model, I tried 10 iterations of each of the above values of t_delta for delta=1 and 0.1, 20 for 0.1e-3 and 30 for 1e-4. 

What I noticed is that, with small values of t_delta there was no benefit to the overall score after 10 iterations. In fact, with t_delta=1e-4, the score after 10 iterations was just 71 %, where as with t_delta=1. and 0.1, the score was close to 81% after 10 iterations. After nearly 23 epochs, the score increased to 81% with t_delta=1e-4.

On the flip side, increasing it even a small amount more than 1.0 led to slower learning, and in the case of t_delta=2. actually led to gradient explosion and no learning at all.

That is the reason I have reverted the default value of t_delta to 1.0, since it is a good enough value. Users can change it to perform more accurate training perhaps, but making it too low is just slowing down training with no benefits.

keras_contrib/layers/normalization.py
titu1994: See the graph in https://github.com/farizrahman4u/keras-contrib/issues/17. This is comparing Batchnorm mode 0 with Batchrenorm mode 2. Mode 2 output is simply output of x_normed at both training and testing time. It is showing clearly different training and validation curves to the the batchnorm mode 0.

See the graph in this PR, which is comparing Mode 0 batch renorm with mode 0 batchnorm. As you can see, the curve for both is similar. This suggests that the batch renorm is performing similar to batchnorm, all other things kept constant. This is validated by the paper's Figure 1, in which the validation curve of Inception v3 with batchnorm and batchrenorm are completely similar, with batchrenorm getting a small gain in accuracy. 

There is a clear difference between mode 0 and mode 2 behavior. This suggests x_normed does not become equivalent to x_normed_running after r and d are at their max values, and the graph below suggests this is correct behavior with what was observed in the paper.

keras_contrib/layers/normalization.py
titu1994: I tried using different values of t_delta. See my comment at the end of the PR to see the results, but basically, the only difference is with convergence speed. They all end up at similar results, but with very small values of t_delta, it takes far longer to converge to the same result.

On the other hand, having larger values of t_delta causes gradient explosion, like what I saw before I started using the time step calculated d_max and r_max.

keras_contrib/layers/normalization.py
titu1994: Also, the algorithm implemented here is exactly matching the one described in the paper. The only difference is in how the r_max and d_max are updated, since they use many more iterations.

Other than that, the implementation matches Algorithm 1 exactly, and the results conform to what is observed by the paper (at mode 0, both batch norm and batch renorm should behave similarly, with batch renorm giving slight increase in accuracy).

Please tell me, what exactly do you want me to change, and how will that change affect the training graph?

titu1994:
@the-moliver @patyork This implementation is complete now, according to me. Can someone run some final checks on the finished algo against the paper? On my end, the results are always similar to what is described in the paper now.

I am still debating keeping mode=2 for compatibility with batchnorm, but it's basically just an alias for mode=0 for BRn.

keras_contrib/layers/normalization.py
```diff
@@ -1,3 +1,236 @@
...
+            d = (m - self.running_mean) / (self.running_std + self.epsilon)
+            d = K.stop_gradient(K.clip(d, -d_max_val, d_max_val))
+
+            x_normed = ((x_normed_batch * r) + d) * self.gamma + self.beta
+
+            # update r_max and d_max
+            t_val = K.get_value(self.t)
+            r_val = self.r_max_value / (1 + (self.r_max_value - 1) * np.exp(-t_val))
+            d_val = self.d_max_value / (1 + ((self.d_max_value / 1e-3) - 1) * np.exp(-(2 * t_val)))
+            t_val += 1.
```
the-moliver: This should be `t_val += float(self.t_delta)`

keras_contrib/layers/normalization.py
```diff
@@ -1,3 +1,236 @@
...
+        self.mode = mode
+        self.axis = axis
+        self.momentum = momentum
+        self.gamma_regularizer = regularizers.get(gamma_regularizer)
+        self.beta_regularizer = regularizers.get(beta_regularizer)
+        self.initial_weights = weights
+        self.r_max_value = r_max_val
+        self.d_max_value = d_max_val
+        self.t_delta = t_delta
+        if self.mode == 0 or self.mode == 2:
```
the-moliver: I think it's probably best to remove mode 2 and warn the user if they try to use it, since mode 2 in batch renorm behaves very differently to mode 2 in batch norm

keras_contrib/layers/normalization.py
the-moliver: Ok, because it might be generally useful, you could add a `moments` method to both the backends and then call that here as `mean_batch, var = KC.moments(x, reduction_axes)`

keras_contrib/layers/normalization.py
titu1994: I've dropped support for mode 2 in renorm. I think that was a mistake. Of course, the paper suggest the inference algorithm as mode 0, but my previous mode 2 is using the r and d parameters even at test time. 

This leads to even faster convergence and better performance. Also, I can't use the mode 0 output well in semantic segmentation models, whereas mode 2 works very well. 

I am thinking of reverting to earlier version, where mode 0 will perform test time on running statistics, and mode 2 will perform test time on r and d computer for that minibatch.

keras_contrib/layers/normalization.py
titu1994: There is no equivalent to moments in Theano . It would be a simple wrapper over these two functions at best and provide no performance improvements in Theano backend.

the-moliver: Yeah I know, that's fine. It might be a nice convenience function. That's the way a lot of things are when having more than one backend.

patyork: What he's saying is to add a function to the theano backend that accomplishes this; where the TF backend just calls the tf.nn.moments function. It would eliminate the backend-specific code here, and provide future users a method to call directly.

titu1994: Oh. Alright. I wasn't quite understanding what the problem was. Will add it to the backend in some time.

keras_contrib/layers/normalization.py
the-moliver: Well convergence shouldn't change between the two since the estimation code is identical, only inference. Inference _could only_ be different if clipping is happening, but the effect should be minimal. If clipping is happening that suggests r and or d should be set to larger max values.

the-moliver: However, it is possible that due to the strange statistics of text data, clipping may be a good thing. So I'm not necessarily opposed to including it, I just want to make sure it's not a problem that can be resolved with better hyperparameter settings within the context of the original algorithm. Your proposed mode 2 is novel and will behave quite different from mode 2 of batchnorm, so it needs good justification for approval.

titu1994: Hmm it is true that they _should_ be same at inference time, since clipping should be minimal at the end when r and d are max, for some reason, no matter what value I used for t_delta, I could not properly train a Renorm model for semantic segmentation. 

From the very beginning, the dice_score just oscillates between 0.4 and 0.47. The moment I switched to the Mode 2 on the older version (with the std / variance problem fixed), the score leaped as expected from 0.4 to 0.76 in just 8 epochs. After that it rose slowly, to a final value of about 0.84 or so after 100 epochs. The training images were definitely very difficult to train, but it still got a good score considering the model was quite small. 

For reference, with batchnorm mode 0, it was stuck at 0.2-0.34. With mode 2, it rose to a maximum of 0.8293 after 100 epochs. This means that unlike batchnorm mode 0 which at testing time uses moving statistics, BRN is actually working correctly, since at mode 0 (which merges mode 0 and 2 of BN), it is still learning, just not well enough. 

I agree that it is definitely a model technique, not described in the paper. I will leave it to your judgement to decide if it should be included.

keras_contrib/layers/normalization.py
titu1994: Thanks. Forgot to change it.

keras_contrib/layers/normalization.py
titu1994: Hmm, I may have to wait until the DenseNet FCN model PR gets merged first. If I try to push now, it will later cause a merge conflict with the master branch.

keras_contrib/layers/normalization.py
the-moliver: Huh, that is interesting. Have you tried higher values for d_max and r_max? Can you verify that setting them high enough so that no clipping happens, starts hurting performance? This is just to check that everything is behaving as expected.

I've always found mode 2 in the original batchnorm a bit strange, since inference will change depending on how you batch your data, which is undesirable to say the least. I assume it was included previously because some input distributions are so wonky that normalization helps inference more than the non-determinism hurts. And I assume something like that must be going on here too. And perhaps this is a better solution since only extreme values will behave non-deterministically like batchnorm mode 2, while most values will behave deterministically.

titu1994: I always use batch norm mode 2 when doing tasks such as Image inpainting, super resolution or semantic segmentation. I find that not only does it perform better, sometimes with mode 0 it wont even train at all (as in this case, and a long time ago when I was doing super resolution with a GAN).

Also, I set r_max and d_max to 1e8 as a test (I dont want to try np.inf, it may be overkill). Training did not even budge until about the 17th epoch, and even then it rose only a little and then flatlined. So I think everything is working properly in BRN.

tests/keras_contrib/backend/test_backends.py
```diff
@@ -99,6 +100,32 @@ def test_extract(self):
...
                 assert_allclose(zth, ztf, atol=1e-02)
 
+    def test_moments(self):
+        input_shape = (10, 10, 10, 10)
+        x_0 = np.zeros(input_shape)
+        x_1 = np.ones(input_shape)
+        x_random = np.random.random(input_shape)
+
+        th_axes = [0, 2, 3]
+        tf_axes = [0, 1, 2]
```
the-moliver: I don't think these axes lists should be different.

titu1994: Oh I was thinking about theano used with tf dim ordering and tensorflow with th. Should I remove them?

the-moliver: Well your function is compute mean/var on the same `ip` along different dimensions and then comparing them which is probably why tests are failing.

titu1994: Ah no the tests were failing cause of extremely small tolerance. They are passing now.

titu1994: @the-moliver  What about the mode 2 addition? Should I go through with it?

the-moliver: Wait, how can that possibly be? Why should these be the same with different axes numbers? I don't see why the different ordering of axes between tf and th would apply here.

`th_mean, th_var = KCTH.moments(ip_th, 2, keep_dims=keep_dims)`
`tf_mean, tf_var = KCTF.moments(ip_tf, 1, keep_dims=keep_dims)`

the-moliver: Yeah, it seems worth having.

titu1994: Ok. Will push now

titu1994: Ok. Pushed the final Mode 2. Took so long cause I had to make sure everything else was correct as well.

titu1994:
@the-moliver @patyork The PR is now finally ready to be merged

PR #30, opened 2017-02-20
[WIP] densenet.py image segmentation support

ahundt:
uses Fully Convolutional Network (FCN)

### DIFF
--- a/keras_contrib/applications/densenet.py
+++ b/keras_contrib/applications/densenet.py
@@ -57,7 +57,7 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_dense_block_layers=-
             growth_rate: number of filters to add per dense block
             nb_filter: initial number of filters. -1 indicates initial
                 number of filters is 2 * growth_rate
-            nb_dense_block_layers: number of layers in each dense block. 
+            nb_dense_block_layers: number of layers in each dense block.
                    Can be an -1, a positive integer or a list
                    If -1, it computes the nb_layer from depth.
                    If positive integer, a set number of layers per dense block.
@@ -84,8 +84,8 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_dense_block_layers=-
             classes: optional number of classes to classify images
                 into, only to be specified if `include_top` is True, and
                 if no `weights` argument is specified.
-            segmentation: False default produces single label per image with 
-                DenseNet structure, True produces segmentation with single 
+            segmentation: False default produces single label per image with
+                DenseNet structure, True produces segmentation with single
                 label per pixel with FCN Densenet structure.
 
         # Returns
@@ -116,12 +116,12 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_dense_block_layers=-
         else:
             img_input = input_tensor
 
-    x = __create_dense_net(classes, img_input, include_top=include_top, depth=depth, 
-                           nb_dense_block=nb_dense_block, growth_rate=growth_rate, 
-                        nb_filter=nb_filter, nb_dense_block_layers=nb_dense_block_layers,
-                        bottleneck=bottleneck, reduction=reduction, 
-                        dropout_rate=dropout_rate, weight_decay=weight_decay, 
-                        segmentation=segmentation, input_shape=input_shape)
+    x = __create_dense_net(classes, img_input, include_top=include_top, depth=depth,
+                           nb_dense_block=nb_dense_block, growth_rate=growth_rate,
+                           nb_filter=nb_filter, nb_dense_block_layers=nb_dense_block_layers,
+                           bottleneck=bottleneck, reduction=reduction,
+                           dropout_rate=dropout_rate, weight_decay=weight_decay,
+                           segmentation=segmentation, input_shape=input_shape)
 
     # Ensure that the model takes into account
     # any potential predecessors of `input_tensor`.
@@ -198,7 +198,9 @@ def __conv_block(ip, nb_filter, bottleneck=False, dropout_rate=None, weight_deca
     x = Activation('relu')(x)
 
     if bottleneck:
-        inter_channel = nb_filter * 4  # Obtained from https://github.com/liuzhuang13/DenseNet/blob/master/densenet.lua
+        # Obtained from
+        # https://github.com/liuzhuang13/DenseNet/blob/master/densenet.lua
+        inter_channel = nb_filter * 4
 
         x = Convolution2D(inter_channel, 1, 1, init='he_uniform', border_mode='same', bias=False,
                           W_regularizer=l2(weight_decay))(x)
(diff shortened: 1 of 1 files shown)


titu1994:
@ahundt This seems to merge my DenseNet model with my other repo for FCN DenseNet. Have you tested it to see that it correctly builds and loads the weights for normal DenseNet models?

ahundt:
@titu1994 yes I've test it with normal densenet models

ahundt:
I know you imagined two separate classes, but when I started making the changes I realized that the code overlap is very substantial and items like the list of layer thicknesses could be used to extend the parameterization available to the original densenet model as well, so I decided to make it so there is simply a `segmentation=True` or `segmentation=False` flag, where it defaults to the original behavior.

titu1994:
@ahundt Cool that's pretty neat work merging the two model builder scripts. 

On another note, I see you went with Deconvolution networks instead of UpSampling 2D or SubpixelConvollution Layers like in my repo. Deconvolution layers tend to produce blocky output images, especially in domains like image inpainting and single image super resolution (I know this is for segmentation, but it FCNs can be easily applied to the above tasks as well).

Since UpSampling 2D seems to be a better at handling those blocky patches, I think an optional flag should be added to support UpSampling2D layers. Later when the SubpixelUpsampling layer is corrected to work with tensorflow, we can add that as an option as well. Reasoning is from the post http://distill.pub/2016/deconv-checkerboard/

ahundt:
@titu1994 yeah I think there should be a parameter with a few different upsampling options, just I'd like to make those changes separate pull requests that create actual upsampling keras layers for that purpose (like the atrous link above) and add the options in those future pulls.

titu1994:
@ahundt Oh that's useful as well. We definitely need more upsampling layers. The simple UpSampling2D is good, but dated. I'll look to see if I can get SubpixelConvolutionLayer working with TF properly and then add a PR. 

On a side note, what do the tests for upsampling layers even look like?

ahundt:
@titu1994 we can move the upsampling conversation to #31

keras_contrib/applications/densenet.py
```diff
@@ -301,25 +373,93 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
...
+        for block_idx in range(nb_dense_block):
+            x, nb_filter = __dense_block(x, nb_dense_block_layers[block_idx], nb_filter, growth_rate, bottleneck=bottleneck,
+                                         dropout_rate=dropout_rate, weight_decay=weight_decay)
+            # Skip connection
+            x = merge([x, skip_connection], mode='concat',
+                      concat_axis=concat_axis)
+            skip_list.append(x)
+
+            # add transition_block
+            x = transition_down_block(x, nb_filter, compression=compression, dropout_rate=dropout_rate,
```
titu1994: This should be __transition_down_block

keras_contrib/applications/densenet.py
```diff
@@ -301,25 +373,93 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
...
+                                      weight_decay=weight_decay)
+            # this is calculated inside transition_down_block
+            nb_filter = int(nb_filter * compression)
+
+            # Preserve transition for next skip connection after dense
+            skip_connection = x
+
+        # The last dense_block does not have a transition_down_block
+        x, nb_filter = dense_block(x, final_nb_layer, nb_filter, growth_rate, bottleneck=bottleneck,
```
titu1994: This should be __dense_block

keras_contrib/applications/densenet.py
```diff
@@ -14,7 +16,7 @@
 
 from keras.models import Model
 from keras.layers.core import Dense, Dropout, Activation
```
titu1994: Haven't imported Reshape layer

keras_contrib/applications/densenet.py
```diff
@@ -301,25 +373,93 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
...
+
+            # Preserve transition for next skip connection after dense
+            skip_connection = x
+
+        # The last dense_block does not have a transition_down_block
+        x, nb_filter = dense_block(x, final_nb_layer, nb_filter, growth_rate, bottleneck=bottleneck,
+                                   dropout_rate=dropout_rate, weight_decay=weight_decay)
+
+        if K.image_dim_ordering() == 'th':
+            out_shape = [batch_size, nb_filter, rows // 16, cols // 16]
```
titu1994: Batch size is unknown in model definition. That is why I am avoiding usage of Deconvolution2D Layers. Wait for Keras 2 to arrive, in which output shape is auto inferred.

keras_contrib/applications/densenet.py
```diff
@@ -301,25 +373,93 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
...
+        else:
+            out_shape = [batch_size, rows // 16, cols // 16, nb_filter]
```
titu1994: Same as above

keras_contrib/applications/densenet.py
```diff
@@ -301,25 +373,93 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
...
+
+        # Add dense blocks and transition up block
+        for block_idx in range(nb_dense_block):
+            if K.image_dim_ordering() != 'th':
+                out_shape[3] = nb_filter
+
+            x = __transition_up_block(
+                x, nb_filters=nb_filter, type=upscaling_type, output_shape=out_shape)
```
titu1994: You removed the upscaling_type parameter in my original model definition, but still call it here.

patyork:
>How did this code even pass the tests if the code itself was riddled with bugs, I have no idea.

There aren't any tests included with this PR, so all the code had to do was pass PEP and not break anything else..

@ahundt As with everything, always run your code before sharing it, and especially before making a PR. These are similar issues as with your other PR I glanced at.

ahundt:
How would you recommend an applications test work, anything aside from creating the model in several different configurations? I believe actual training won't be appropriate for a unit test.

titu1994:
There's no need for a model hiding test. But ateast entire that the model's that it built are correct. 

Correctness would mean - correct number of parameters, correct loading of weights without errors, no bugs in the code. The general checks one does before making code publicly consumable by the general developers community.

titu1994:
@ahundt If you're not in a hurry, give me till Friday. I have a test on Thursday, but once that's done, I can work on merging my two repos properly into a general framework for building DenseNets and DenseNet FCNs.

titu1994:
@ahundt Added a PR for a working DenseNet FCN

ahundt:
Sorry about that, I've contributed to big projects before and it isn't unusual to create a pull request to get early feedback on possible changes before everything is done and time is wasted on code or a design that could have been better, but again it should have been marked WIP and said so.

I'll close this in favor of #34 and add comments to that

titu1994:
@ahundt Oh I didn't know about that. Generally, I don't publish any code that I haven't tested thoroughly first so it was a bit weird for me.

https://github.com/farizrahman4u/keras-contrib/pull/34 manages to merge densenet and densenet fcns neatly, albeit I am still looking into tests for SubpixelUpscaling and depth_to_scale backend functions. Other layers are pretty stable.

PR #33, opened 2017-02-21
[WIP] Add FCN DenseNets

titu1994:
Adds DenseNet FCN model builder script, depth_to_space operations in backend and SubpixelConvolutionUpsampling layer.

Todo:

1) See if we can test depth_to_space (I don't know how to do this. But I have tested the Subpixel convolutional layer which uses this, and output is correct. Does it imply that the operations in depth_to_space is correct as well?
2) See if we can somehow test SubpixelUsampling layer. The problem is it does not take in an input size, it simply receives the output of the previous layer and shuffles it to obtain a larger size in exchange for smaller number of kernels. Implementation should be correct, but tests need to be designed to work with an upsampling technique which is this unique.

### DIFF
--- a/keras_contrib/applications/densenet_fcn.py
+++ b/keras_contrib/applications/densenet_fcn.py
@@ -11,10 +11,9 @@ from __future__ import absolute_import
 from __future__ import division
 
 from keras.models import Model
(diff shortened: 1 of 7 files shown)


titu1994:
Closing cause unrelated changes cropped up in this PR

PR #34, opened 2017-02-21
Added DenseNet FCN

titu1994:
Adds DenseNet FCN model builder script, depth_to_space operations in backend and SubpixelConvolutionUpsampling layer.

Todo:

1) See if we can test depth_to_space (I don't know how to do this). But I have used the Subpixel convolutional layer for upsampling tasks which uses this, and output is correct. Does it imply that the operations in depth_to_space is correct as well?

2) See if we can somehow test SubpixelUsampling layer. The problem is it does not take in an input size, it simply receives the output of the previous layer and shuffles it to obtain a larger size in exchange for smaller number of kernels. Implementation should be correct, but tests need to be designed to work with an upsampling technique which is this unique.


keras_contrib/applications/densenet.py
```diff
@@ -13,8 +13,8 @@
 import warnings
 
 from keras.models import Model
-from keras.layers.core import Dense, Dropout, Activation
-from keras.layers.convolutional import Convolution2D
+from keras.layers.core import Dense, Dropout, Activation, Reshape
```
ahundt: replace appropriate top comment lines with this:
```
 - [Densely Connected Convolutional Networks](https://arxiv.org/pdf/1608.06993.pdf) for image labeling
  - [The One Hundred Layers Tiramisu: Fully Convolutional DenseNets for Semantic Segmentation](https://arxiv.org/pdf/1611.09326.pdf)
     for Fully Convolutional Network (FCN) based image segmentation
```

keras_contrib/applications/densenet.py
```diff
@@ -164,6 +166,114 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_filter=16,
     return model
 
 
+def DenseNetFCN(nb_dense_block=5, growth_rate=12, nb_filter=16, nb_layers_per_block=4,
```
ahundt: Original densenet can also be appropriately parameterized with nb_dense_block_layers this is currently lost from #30.

keras_contrib/applications/densenet.py
```diff
@@ -275,7 +419,7 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
         dropout_rate: dropout rate
         weight_decay: weight decay
 
-    Returns: keras tensor with nb_layers of __conv_block appended
+    Returns: keras tensor with nb_layers of conv_block appended
```
ahundt: mistake?

keras_contrib/applications/densenet.py
```diff
@@ -323,3 +467,146 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
         x = Dense(nb_classes, activation='softmax', W_regularizer=l2(weight_decay), b_regularizer=l2(weight_decay))(x)
 
     return x
+
+
+def __create_fcn_dense_net(nb_classes, img_input, include_top, nb_dense_block=5, growth_rate=12,
```
ahundt: could reduce duplicate code by merging with __create_dense_net?

keras_contrib/applications/densenet.py
```diff
@@ -323,3 +467,146 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
...
+
+        x = merge([x, skip_list.pop()], mode='concat', concat_axis=concat_axis)
+
+        x, nb_filter = __dense_block(x, nb_layers[-block_idx], nb_filter, growth_rate, bottleneck=bottleneck,
+                                     dropout_rate=dropout_rate, weight_decay=weight_decay)
+
+    x = Convolution2D(nb_classes, 1, 1, activation='linear', border_mode='same', W_regularizer=l2(weight_decay),
+                      bias=False)(x)
+
+    if K.image_dim_ordering() == 'th':
```
ahundt: need option to include/exclude top

keras_contrib/applications/densenet.py
```diff
@@ -323,3 +467,146 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
...
+
+    # compute initial nb_filter if -1, else accept users initial nb_filter
+    if nb_filter <= 0:
+        nb_filter = 2 * growth_rate
+
+    # compute compression factor
+    compression = 1.0 - reduction
+
+    # Initial convolution
+    x = Convolution2D(48, 3, 3, init="he_uniform", border_mode="same", name="initial_conv2D", bias=False,
```
ahundt: 48 should be nb_filter parameter

keras_contrib/applications/densenet.py
```diff
@@ -323,3 +467,146 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
         x = Dense(nb_classes, activation='softmax', W_regularizer=l2(weight_decay), b_regularizer=l2(weight_decay))(x)
 
     return x
+
+
+def __create_fcn_dense_net(nb_classes, img_input, include_top, nb_dense_block=5, growth_rate=12,
+                           nb_filter=-1, bottleneck=False, reduction=0.0, dropout_rate=None, weight_decay=1E-4,
+                           nb_layers_per_block=4, nb_upsampling_conv=128, upsampling_type='upsampling',
```
ahundt: rename to `nb_dense_block_layers` so naming matches `nb_dense_block` variable, should be consistently plural/not plural so perhaps `nb_dense_blocks`. 

I suggest the standard densenet should also accept `nb_dense_block_layers` in both integer and list form as well.

keras_contrib/applications/densenet.py
```diff
@@ -323,3 +467,146 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
...
+
+    # Add dense blocks and transition up block
+    for block_idx in range(nb_dense_block):
+        if K.image_dim_ordering() != 'th':
+            out_shape[3] = nb_filter
+
+        x = __transition_up_block(x, nb_filters=nb_filter, type=upsampling_type, output_shape=out_shape)
+
+        if K.image_dim_ordering() == 'th':
+            out_shape[2] *= 2
```
ahundt: maybe parameterize this scaling rate? (in both densenet and densenetfcn)

keras_contrib/applications/densenet.py
```diff
@@ -258,6 +370,38 @@ def __dense_block(x, nb_layers, nb_filter, growth_rate, bottleneck=False, dropou
...
+        x = AtrousConvolution2D(nb_filters, 3, 3, activation="relu", W_regularizer=l2(weight_decay),
+                                bias=False, atrous_rate=(2, 2))(ip)
+    else:
+        x = Deconvolution2D(nb_filters, 3, 3, output_shape, activation='relu', border_mode='same',
+                            subsample=(2, 2))(ip)
+
+    return x
+
+
 def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_block=3, growth_rate=12, nb_filter=-1,
```
ahundt: would be nice to specify number of layers per dense block so it can be parameterized as an integer or a list

keras_contrib/applications/densenet.py
titu1994: I didn't add it specifically bacause the original paper uses depth parameter to achieve this.

In FCN Densenet, there is no depth parameter. Instead you control depth with the two parameters nb_dense_blocks and nb_layers_per_block.

keras_contrib/applications/densenet.py
```diff
@@ -258,6 +368,38 @@ def __dense_block(x, nb_layers, nb_filter, growth_rate, bottleneck=False, dropou
...
+        x = UpSampling2D()(ip)
+    elif type == 'subpixel':
+        x = Convolution2D(nb_filters, 3, 3, activation="relu", border_mode='same', W_regularizer=l2(weight_decay),
+                          bias=False)(ip)
+        x = SubPixelUpscaling(r=2, channels=int(nb_filters // 4))(x)
+        x = Convolution2D(nb_filters, 3, 3, activation="relu", border_mode='same', W_regularizer=l2(weight_decay),
+                          bias=False)(x)
+    elif type == 'atrous':
+        # waiting on https://github.com/fchollet/keras/issues/4018
+        x = AtrousConvolution2D(nb_filters, 3, 3, activation="relu", W_regularizer=l2(weight_decay),
```
titu1994:
```
Additionally the original DenseNet (not the tiramisu U-net version) can be updated to become FCN through atrous convolutions (non-transpose) and loading the original weights. 
``` 
How does that even work? The atrous version would perform completely different operations on the network. Loading the weights of normal densenet, even if it has the same number of layers would give completely random output.

keras_contrib/applications/densenet.py
titu1994: See comment above. The original densenet does not allow modification to this parameter, so i did not include it either.

keras_contrib/applications/densenet.py
ahundt: Any original paper can of course be improved upon. :-) Specifying the number of dense block layers separately from the number of dense blocks is simply a better parameterization. It is also possible to support the original parameterization and the newer one.

keras_contrib/applications/densenet.py
titu1994: No. I use __ in all the methods to make them hidden to the end user. However, when referring to code blocks, there is no need to mention it in private form.

keras_contrib/applications/densenet.py
titu1994: No. It is better to keep it separate. This is what I meant by merging them would cause problems. It is cleaner to keep two separate model building logic completely separate.

keras_contrib/applications/densenet.py
titu1994: This is the initial convolution kernel. As suggested in the Tiramisu paper, it is always 48. Using the variable over here means the user can alter the initial number of convolutions.

keras_contrib/applications/densenet.py
titu1994: DenseNet does not use scale parameter at all. 
DenseNet FCN uses 2x scaling because it first performs 2 ^ nb_dense_block downscales and then 2 ^ nb_dense_block upscales. 

All papers I have seen mention 2x scaling to be important. Any larger and it looses too much information when downscaling, and cannot recover it when upscaling.

keras_contrib/applications/densenet.py
titu1994: It is not possible to define a FCN without a "top" layer though. "top" to a FCN is the final class wise pixel loss which will be used to train the FCN.

keras_contrib/layers/convolutional.py
```diff
@@ -227,3 +227,31 @@ def get_config(self):
 Deconv3D = Deconvolution3D
 get_custom_objects().update({"Deconvolution3D": Deconvolution3D})
 get_custom_objects().update({"Deconv3D": Deconv3D})
+
+
+class SubPixelUpscaling(Layer):
```
titu1994: No it is tensorflow and theano now. It directly calls tensorflow's depth_to_scale method in the tensorflow backend.

keras_contrib/backend/tensorflow_backend.py
```diff
@@ -99,3 +99,9 @@ def extract_image_patches(X, ksizes, ssizes, border_mode="same", dim_ordering="t
     if dim_ordering == "tf":
         patches = KTF.permute_dimensions(patches, [0, 1, 2, 4, 5, 3])
     return patches
+
+
+def depth_to_space(input, scale, **kwargs):
+    ''' Uses phase shift algorithm to convert channels/depth for spacial resolution '''
```
titu1994: Tensorflow calls it depth_to_space. Therefore I have named it similarly.

keras_contrib/applications/densenet.py
titu1994: Well, I guess if it defaults to original densenet structure, no harm no foul. Will implement this.

titu1994: Done.

keras_contrib/applications/densenet.py
titu1994: Done.

keras_contrib/applications/densenet.py
titu1994: Done.

keras_contrib/applications/densenet.py
titu1994: In any case, I have added a warning to state that this does not currently work, and that it is falling back to UpSampling layer.

keras_contrib/backend/tensorflow_backend.py
ahundt: I meant that in the comment on this line you misspelled the word "spatial" as "spacial"

keras_contrib/applications/densenet.py
ahundt: Yes, that is correct if your problem domain is strictly segmentation for evaluating an FCN, but I'm going to incorporate additional layers because FCN is merely a component of my larger network. If you prefer I can make that a future pull request after this is merged.

keras_contrib/applications/densenet.py
ahundt: okay that sounds like a good for now!

> How does that even work?

See Multi-Scale Context Aggregation by Dilated Convolutions
https://arxiv.org/abs/1511.07122

Also some of the materials myself and patyork links to in https://github.com/fchollet/keras/issues/4018 have info specific to this.

keras_contrib/applications/densenet.py
ahundt: cool!

keras_contrib/applications/densenet.py
ahundt: sounds good

keras_contrib/applications/densenet.py
ahundt: fair enough

keras_contrib/applications/densenet.py
ahundt: thanks

keras_contrib/applications/densenet.py
ahundt: Okay if the value differs from `nb_filter`, can this be a separate hyperparameter `initial_conv_nb_filter` that defaults to 48?

keras_contrib/applications/densenet.py
ahundt: good reasons, thanks

keras_contrib/layers/convolutional.py
ahundt: cool!

keras_contrib/applications/densenet.py
ahundt: great!

ahundt:
Once these are fixed I'd say LGTM:
- `spatial` vs `spacial` spelling mistake
- add that extra "top" option (see comment)
- first layer filter option `initial_conv_nb_filter`

keras_contrib/applications/densenet.py
titu1994: So which final layers should I remove? The Convolution 1x1 and the reshape and activation?

keras_contrib/backend/tensorflow_backend.py
titu1994: Ah sorry. Was in the middle of class so I didn't see it properly.

titu1994:
Corrected and fixed all the changes requested by @ahundt

keras_contrib/applications/densenet.py
titu1994: Made the changes. If include_top is false, will not add the Conv 1x1, Reshape and Activation layers at the end.

If I may ask, what problem domain requires DenseNet FCN but even more layers after that? Seems like either Super Resolution or Inpainting.

ahundt:
LGTM, @patyork?

cepera-ang:
It's look like that up-sampling path here implemented in different way than proposed in the "Tiramisu" paper (page 4): 

>  The upsampled fea-
> ture maps are then concatenated to the ones coming from
> the skip connection to form the input of a new dense block.
> _Since the upsampling path increases the feature maps spa-
> tial resolution, the linear growth in the number of features
> would be too memory demanding, especially for the full
> resolution features in the pre-softmax layer._

> In order to overcome this limitation, the input of a dense
> block is not concatenated with its output. Thus, the trans-
> posed convolution is applied only to the feature maps ob-
> tained by the last dense block and **not to all feature maps
> concatenated so far.**

So, dense blocks in the up-sample path should be created without concatenation of its input and output (as it is currently done). 

Somehow related to previous, it looks like number of convolutions in upsample path (`nb_filter`) is calculated in wrong way. Currently, it growth after each `__dense_block` function call. However, number of convolutions in TransitionUp block should be equal to the number of features from the previous dense block (`=nb_layers*growth_rate`), so TU block doesn't change the number of features, just upsamples it. 

- [The One Hundred Layers Tiramisu: Fully Convolutional DenseNets for Semantic Segmentation](https://arxiv.org/abs/1611.09326)

titu1994:
Hmm, I will fix this in two days. I'm a little busy right now with midsems.

the-moliver:
Also, there need to be tests of the new backend functions.

titu1994:
I did add them. They check if Theano and TF and calculating same mean and var

keras_contrib/applications/densenet.py
titu1994: @ahundt Thanks for the info!

titu1994:
@ahundt @the-moliver I'm quite stuck on how I should test the SubPixelUpsampling layer and the new backend functions.

The backend functions first : 
- Require a tensor with 2 * k filters, so a Convolution layer must precede a call to this function.

The subpixel layer also has the same problem - it needs a convolution layer to precede it.

How do you recommend I go about this?

ahundt:
I a super basic test is to instantiate it and make sure it doesn't fail, and run it and check that the dimensions are correct. Basically [do what they do for deconvolution](https://github.com/fchollet/keras/blob/master/tests/keras/layers/test_convolutional.py#L138) and that should be enough for them to merge the changes.

titu1994:
@ahundt @patyork Finished implementation with layer test (which also tests the backend function in one go).

titu1994:
@patyork Found the bug. It's trying to use Theano with dim ordering "TF", but depth to scale only works with TH dim ordering with Theano, and TF dim ordering with TF. I'm going to override the default settings in the code to fix the bugs.

titu1994:
@patyork @ahundt Finished with implementation. Added an assertion to ensure that correct dim ordering is used with correct backend when using depth to space methods.

The implementation is closely in line with the paper and correctly calculates the number of parameters for each model. 

Ready to merge once you go review it.

titu1994:
@kemaswill @patyork The implementation is finished. It can be merged now.

keras_contrib/layers/convolutional.py
```diff
@@ -456,6 +458,101 @@ def get_config(self):
...
+        super(SubPixelUpscaling, self).__init__(**kwargs)
+
+        self.scale_factor = scale_factor
+        self.dim_ordering = dim_ordering
+
+        if self.dim_ordering == 'default':
+            self.dim_ordering = K.image_dim_ordering()
+
+        if (K.backend() == 'theano' and self.dim_ordering == 'tf') or \
+                (K.backend() == 'tensorflow' and self.dim_ordering == 'th'):
```
patyork: Is there a reason you can't dimshuffle, a la Keras, [before ](https://github.com/fchollet/keras/blob/master/keras/backend/theano_backend.py#L1501-L1502)and [after](https://github.com/fchollet/keras/blob/master/keras/backend/theano_backend.py#L1534-L1535)?

titu1994: Hmm let me try it. When I did it manually, it crashed with tensor errors which I couldn't debug properly.

titu1994: @patyork Fixed it. Thanks for the info on backend handling.

titu1994:
@patyork @the-moliver @kemaswill  Can this PR be reviewed and merged? I have finished all the requested changes.

keras_contrib/applications/densenet.py
```diff
@@ -285,10 +451,25 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
         assert reduction <= 1.0 and reduction > 0.0, "reduction value must lie between 0.0 and 1.0"
 
     # layers in each dense block
-    nb_layers = int((depth - 4) / 3)
+    # layers in each dense block
```
ahundt: comment repeats twice

keras_contrib/backend/theano_backend.py
```diff
@@ -117,6 +121,24 @@ def extract_image_patches(X, ksizes, strides, border_mode="valid", dim_ordering=
     return patches
 
 
+def depth_to_space(input, scale):
```
ahundt: depth_to_space might need its own unit test? Although is called in the pixel upscaling test.

titu1994: Cant test it on its own. That's why I am using it inside the SubPixelUpscaling layer test to test it at the same time. 

The fact that the theano implementation returns the same result as the tensorflow wrapper shows that is is correct.

keras_contrib/applications/densenet.py
titu1994: Fixed

titu1994:
The test failure is unrelated to this PR. Closing and re-opening the pr to force tests again.

titu1994:
As again, test failures are not due to the PR. There is some problem with requests in the tests.

tboquet:
@titu1994 @ahundt  LGTM, I restarted the build and everything is green.

titu1994:
@patyork @tboquet Could you merge this please ? I'm using it in one of my projects and it would be useful to have it available in the master branch of contrib

ahundt:
I second this request

patyork:
Build still passes.

PR #36, opened 2017-02-22
Cosine Normalized Dense Layer

the-moliver:
Implementation of layer from:
Cosine Normalization: Using Cosine Similarity Instead of Dot Product in Neural Networks
https://arxiv.org/abs/1702.05870

In my testing it definitely speeds convergence and helps generalization compared to regular dense layers.

### DIFF
--- a/keras_contrib/layers/core.py
+++ b/keras_contrib/layers/core.py
@@ -44,6 +44,8 @@ class CosineDense(Layer):
         # after the first layer, you don't need to specify
         # the size of the input anymore:
         model.add(CosineDense(32))
+
+        **Note that a regular Dense layer may work better as the final layer
     ```
 
     # Arguments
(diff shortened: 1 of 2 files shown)


the-moliver:
Merging now. Just a cautionary note, as stated in the docs, don't use it for the last layer. Kinda like how you don't use batchnorm on the outputs of the last layer of a network.

notwa:
deferring the division to the output is faster, but i haven't tested it thoroughly. maybe it doesn't work in the 3D case.
```python
    def call(self, x, mask=None):
        if self.bias:
            xnorm = K.sqrt(K.sum(K.square(x), axis=-1, keepdims=True) + 1 + K.epsilon())
            Wnorm = K.sqrt(K.sum(K.square(self.W), axis=0) + K.square(self.b) + K.epsilon())
            output = (K.dot(x, self.W) + self.b) / (xnorm * Wnorm)
        else:
            xnorm = K.sqrt(K.sum(K.square(x), axis=-1, keepdims=True) + K.epsilon())
            Wnorm = K.sqrt(K.sum(K.square(self.W), axis=0) + K.epsilon())
            output = K.dot(x, self.W) / (xnorm * Wnorm)
        return self.activation(output)
```
thanks for your implementation, by the way. i figured biases would be handled this way, but i wasn't entirely confident.

the-moliver:
Thanks, good to know! Will update both.

PR #37, opened 2017-02-23
Cosine Normalized Conv2D

the-moliver:
A Conv2D layer with cosine normalization to finish off the implementation of:
Cosine Normalization: Using Cosine Similarity Instead of Dot Product in Neural Networks
https://arxiv.org/pdf/1702.05870.pdf


tboquet:
Nice addition, it seems possible to write a wrapper, what do you think?

patyork:
@tboquet Do you mean a wrapper as in `TimeDistributed`?

the-moliver:
That would be cool, but potentially tricky or hacky. I'll look into it.

tboquet:
I'm trying to compare the layer to the classic one. 
(I also tried `CosineDense`)
It doesn't seem so good 😞 . I'm maybe doing something wrong. Have you tried it on MNIST?

Also this seems weird:

> The learning rate of baseline was 0.1, cosine normalization 10, batch normalization 5, weight normalization 5, max normalization 1.

tboquet:
My mistake I was a little bit tired I switched the names of the 2 scripts. I was searching in the paper and suspected that they were too generous on the learning rates.
Everything looks good to me 😄.

tboquet:
LGTM. @patyork @kemaswill any other comments?

keras_contrib/layers/convolutional.py
```diff
@@ -227,3 +228,234 @@ def get_config(self):
 Deconv3D = Deconvolution3D
 get_custom_objects().update({"Deconvolution3D": Deconvolution3D})
 get_custom_objects().update({"Deconv3D": Deconv3D})
+
+
+class CosineConvolution2D(Layer):
+    """Cosine Normalized Convolution operator for filtering windows of two-dimensional inputs.
+    Cosine Normalization: Using Cosine Similarity Instead of Dot Product in Neural Networks
```
kemaswill: Better to move the references to the end of these comments, after the output shape, like https://github.com/fchollet/keras/blob/master/keras/layers/convolutional.py

titu1994:
@the-moliver If I'm not mistaken, Cosine Normalization can be applied like Batch Norm right? So instead of a wrapper, how about a layer like BatchNormalization or BatchRenormalization? 

If I understood the paper correctly, it works on pre-activations, so it should be suitable as a replacement for Batchnorm as well right?

PR #46, opened 2017-03-16
densenet dilation for segmentation

ahundt:
Work in progress for the use of atrous convolution to enable segmentation with the standard DenseNet without a U-net structure. By passing `dilation_rate=2` and `pooling=None`, the resulting network will be full resolution with a receptive field equivalent to the original (and still default) version with average pooling. Then the same 2D activation top as DenseNetFCN can be applied for segmentation results. [Multi-Scale Context Aggregation by Dilated Convolutions](https://arxiv.org/abs/1511.07122) is a good reference on atrous convolutions as is [A guide to convolution arithmetic for deep learning](https://arxiv.org/abs/1603.07285) and its corresponding [conv_arithmetic repository](https://github.com/vdumoulin/conv_arithmetic).

TODO before merging:
 - [x] create instance of atrous segmentation DenseNet
 - [x] train on coco or pascal voc for comparison against baseline
 - [x] add basic test

@titu1994 may be interested as the main author of the DenseNet code.


titu1994:
@ahundt Nice work! I will be a bit busy the next few days due to midsems, but you should carry on. I will read up on the papers in the meantime. 

On that note, perhaps add the paper to the top of the document referenced and add some documentation with example usage in the docstring of the DenseNet class, as people may inadvertently be confused as to why DenseNet has dilation parameters when it normally doesn't use AtrousConvolutions.

keras_contrib/applications/densenet.py
```diff
@@ -113,8 +118,9 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_filter=16, nb_layers
             img_input = input_tensor
```
titu1994: Add a check to see if pooling is one of ['avg', 'max', None], and raise a ValueException otherwise

keras_contrib/applications/densenet.py
```diff
@@ -331,6 +338,9 @@ def __transition_block(ip, nb_filter, compression=1.0, dropout_rate=None, weight
                     in the transition block.
         dropout_rate: dropout rate
         weight_decay: weight decay factor
+        dilation_rate: an integer or tuple/list of 2 integers, specifying the dilation rate to
+        use for dilated convolution. Can be a single integer to specify the same value for
+        all spatial dimensions.
```
titu1994: Add docstring explaining pooling parameters

keras_contrib/applications/densenet.py
```diff
@@ -341,10 +351,14 @@ def __transition_block(ip, nb_filter, compression=1.0, dropout_rate=None, weight
...
-                      W_regularizer=l2(weight_decay))(x)
+                      W_regularizer=l2(weight_decay), dilation_rate=dilation_rate)(x)
     if dropout_rate:
         x = Dropout(dropout_rate)(x)
-    x = AveragePooling2D((2, 2), strides=(2, 2))(x)
+
+    if pooling == "avg":
+        x = AveragePooling2D((2, 2), strides=(2, 2))(x)
+    elif pooling == "max":
+        x = MaxPooling2D((2, 2), strides=(2, 2))(x)
```
titu1994: Handle pooling=None case by removing elif and instead use else: (default to average pooling by using if pooling == 'max')

ahundt: There should be neither `max` nor `avg` pooling in the `None` case, as that changes the dimensions and will affect the resolution of the segmentation output.

patyork:
Please don't open a Pull Request unless the functionality is ready to be merged, especially in a repository where multiple people have merge abilities.

ahundt:
@patyork what's your recommended procedure for early feedback then to ensure ongoing work isn't duplicated or off the rails in terms of approach?

patyork:
It's in CONTRIBUTING.md and is the same as for Keras: Open an issue to explain and gauge interest in/need for/feasibility of the idea and flesh out the scope of the feature; if the idea passes muster (for contrib, it most likely will) fork the repo, make a branch, implement the feature fully per any discussion in the relevant issue; open a pull request referencing the issue; make any changes as requested by reviewers; feature merged.

This is as opposed to everyone opening a PR that isn't ready to merge and: Reviewers auditing broken code, reviewing the same code many times as new "parts" get added or fixed, reviewers not knowing if something is ready for merge/review, poorly thought-out PRs that end up having 20-30 commits on which Travis runs each time, maintainers getting emails for each PR when they have no action item on it, someone spending 10 hours implementing a feature that ends up not being merged at all (bad code, unneeded or duplicate feature) or that upon review could have been done in a different way in 5 minutes, etc, etc, etc. In other words, there's good reason why CONTRIBUTING.md has already spelled all of this out.

-----Original Message-----

ahundt:
Sorry if I've wasted your time, I know as a long time Keras contributor and with Keras being so popular you must have your inbox flooded. Is there a mistake in my understanding that github designed pull requests and issues to be essentially identical modulo associated commits? I find the travis runs are actually quite useful. 

I can follow this procedure in the future if I'm discussing something specific with other contributors:
 - Create an issue & a branch on my fork
 - link to my branch in the issue until a pull request is ready
 - submit pull request

It does seem slightly redundant to me and there will be two #s associated with changes and items to close, but I will live. :-)

I hope you'll understand when also I mention that the contributing.md isn't yet as clear as it could be on this point:

> If your PR introduces a change in functionality, make sure you start by opening an issue to discuss whether the change should be made, and how to handle it. This will save you from having your PR closed down the road! Of course, if your PR is a simple bug fix, you don't need to do that.

This could be read as creating issues for changes in existing, supported functionality which would of course have a much higher bar for acceptance, especially if existing code could break.

One option that may be helpful for you, in some cases consider using the unsubscribe button and devs would likely @  mention you when attention is needed? I've been using unsubscribe for some items and it isn't perfect but it does reduce the noise. Maybe we need to add a new Github issue and PR importance classifier as a Keras example and chrome extension, LOL!

patyork:
Notifications are just a side issue. The main points:

- A *Pull Request* is a request for a merge. Opening a PR that you aren't ready to have merged is fundamentally the opposite of a pull request; it's a *don't pull* request which is the same thing as (if not worse than) not making a request at all.
- The Issue+Pull Request pairing/metaphor is about separating *documentation* from *technical discussion*. The issue (it's a poor name in github, IMO; "Discussion" or "Tickets" would be better) should document the **why** of a Pull Request - description of a bug, the theory/math behind a new feature, general discussion around the item, etc.. The Pull Request should document the **how** - the code itself, perhaps some back-and-forth discussion about why the code is structured the way it is, etc.. A person with no programming/Python experience should be able to read the relevant Issue and understand what the PR's goal was. And, as with any project, having a **why** is important before doing the **how**.

The other points stand, but those are my two main ones. Issues and PRs are separate entities for different purposes, and that's why they exist separately.

ahundt:
Thanks again for the clarification, I plan to follow that approach in the future for `keras` and `keras-contrib`. I did want to add that other projects may choose different procedures and definitions from what's done here and those can also be equally valid/useful in their appropriate context. :-)

Actually, your comments may be worth adding directly to both `CONTRIBUTING.md` files.

ahundt:
@titu1994 This is ready for review now, could you take a look? If you won't be available let me know and I will ping others.

Note that the usage I have here is the way atrous convolutions work, so I've removed it from the FCNDenseNet code in 9665d72, perhaps an upscaling dilation rate of some sort could be added to that but it should be future work and would just add a dilation parameter to the already existing Conv2DTranspose.

To run the code clone my repository fork https://github.com/ahundt/Keras-FCN/tree/densenet_atrous and run the `train.py` script. Note that Pascal VOC 2012 will need to be extracted in:
```
~/datasets/VOC2012/VOCdevkit/VOC2012/
```

An [automated Pascal VOC configuration and download script](https://github.com/ahundt/tf-image-segmentation/blob/ahundt-keras/tf_image_segmentation/recipes/pascal_voc/data_pascal_voc.py) can be found at my fork of [tf-image-segmentation on the ahundt-keras branch](https://github.com/ahundt/tf-image-segmentation/tree/ahundt-keras).

It can also be downloaded by hand:
```
    # original PASCAL VOC 2012
    # wget http://host.robots.ox.ac.uk/pascal/VOC/voc2012/VOCtrainval_11-May-2012.tar # 2 GB
```

While I hope to eventually directly incorporate an FCN training script into keras-contrib, I'm waiting on permission from the author of Keras-FCN.

Let me know if any additional information is needed!

keras_contrib/applications/densenet.py
```diff
@@ -230,32 +241,31 @@ def DenseNetFCN(input_shape, nb_dense_block=5, growth_rate=16, nb_layers_per_blo
...
+                         '"upsampling", "deconv", or "subpixel".')
 
     if upsampling_type == 'deconv' and batchsize is None:
         raise ValueError('If "upsampling_type" is deconvoloution, then a fixed '
                          'batch size must be provided in batchsize parameter.')
 
     if input_shape is None:
-        raise ValueError('For fully convolutional models, input shape must be supplied.')
+        raise ValueError(
+            'For fully convolutional models, input shape must be supplied.')
```
titu1994: Can be fit in one line

keras_contrib/applications/densenet.py
```diff
@@ -230,32 +241,31 @@ def DenseNetFCN(input_shape, nb_dense_block=5, growth_rate=16, nb_layers_per_blo
...
     input_shape = _obtain_input_shape(input_shape,
                                       default_size=32,
                                       min_size=16,
-                                      data_format=K.image_dim_ordering(),
-                                      include_top=include_top)
+                                      data_format=K.image_data_format(),
+                                      # If doing segmentation we still include top
+                                      # but _obtain_input_shape only supports
+                                      # labeling, not segmentation networks.
+                                      include_top=False)
```
titu1994: You could wait for my bugfix PR to be merged, it removes this dependency on _obtain_input_shape and uses only those parts of it which make sense to segmentation models. If not, atleast remove the comments inlined with the method arguments and put it above the method call.

keras_contrib/applications/densenet.py
titu1994: The docstring should also explain pooling parameter. Just copy paste from the DenseNet docstring.

keras_contrib/applications/densenet.py
```diff
@@ -368,24 +388,25 @@ def __dense_block(x, nb_layers, nb_filter, growth_rate, bottleneck=False, dropou
...
 
-    concat_axis = 1 if K.image_dim_ordering() == "th" else -1
+    concat_axis = 1 if K.image_data_format() == 'channels_first' else -1
 
     x_list = [x]
 
     for i in range(nb_layers):
-        x = __conv_block(x, growth_rate, bottleneck, dropout_rate, weight_decay)
+        x = __conv_block(x, growth_rate, bottleneck,
+                         dropout_rate, weight_decay)
```
titu1994: Can be fit in one line.

keras_contrib/applications/densenet.py
```diff
@@ -368,24 +388,25 @@ def __dense_block(x, nb_layers, nb_filter, growth_rate, bottleneck=False, dropou
...
         x_list.append(x)
 
-        x1 = concatenate(x_list, axis=concat_axis)
-        #x = merge(x_list, mode='concat', concat_axis=concat_axis)
+        x = merge(x_list, mode='concat', concat_axis=concat_axis)
+        # x = concatenate(x_list, concat_axis)
```
titu1994: Remove commented line for now. Can be updated once a patch for the bug is merged onto Keras master.

keras_contrib/applications/densenet.py
```diff
@@ -404,26 +425,23 @@ def __transition_up_block(ip, nb_filters, type='upsampling', output_shape=None,
...
-    elif type == 'atrous':
-        # waiting on https://github.com/fchollet/keras/issues/4018
-        x = AtrousConvolution2D(nb_filters, 3, 3, activation="relu", W_regularizer=l2(weight_decay),
-                                bias=False, atrous_rate=(2, 2), init='he_uniform')(ip)
+                   use_bias=False, kernel_initializer='he_uniform')(x)
     else:
-        x = Deconvolution2D(nb_filters, 3, 3, output_shape, activation='relu', border_mode='same',
-                            subsample=(2, 2), init='he_uniform')(ip)
+        x = Conv2DTranspose(nb_filters, (3, 3), output_shape, activation='relu', padding='same',
+                            subsample=(2, 2), kernel_initializer='he_uniform')(ip)
```
titu1994: Hmm now that Keras 2 has removed the need for knowing the output shape, the code for Deconv upscaling can be heavily simplified. No more need for the batchsize parameter. Keep it for now since many people (including me) have not shifted to keras 2 yet. Once the bugs from Keras 2 are removed, we can start refactoring this repo to remove support code for keras 1 like batchsize params and output shape computation.

keras_contrib/applications/densenet.py
```diff
@@ -444,13 +462,20 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
...
+        input_shape: Only used for shape inference in fully convolutional networks.
 
     Returns: keras tensor with nb_layers of conv_block appended
     '''
 
-    concat_axis = 1 if K.image_dim_ordering() == "th" else -1
+    concat_axis = 1 if K.image_data_format() == 'channels_first' else -1
 
-    assert (depth - 4) % 3 == 0, "Depth must be 3 N + 4"
+    if depth is not None:
```
titu1994: Can you give me a case when depth is None? As far as I know, depth is never None. Also, I see you have a test case where you build a DenseNet with depth None. How is the None case handled?

keras_contrib/applications/densenet.py
```diff
@@ -501,10 +527,24 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
     x = BatchNormalization(axis=concat_axis, gamma_regularizer=l2(weight_decay),
                            beta_regularizer=l2(weight_decay))(x)
     x = Activation('relu')(x)
-    x = GlobalAveragePooling2D()(x)
+    if pooling is not None:
+        x = GlobalAveragePooling2D()(x)
 
-    if include_top:
-        x = Dense(nb_classes, activation='softmax', kernel_regularizer=l2(weight_decay), bias_regularizer=l2(weight_decay))(x)
+    if include_top and pooling is not None:
```
titu1994: If the argument states to include_top and to pool, you are adding a Dense layer with no pooling. The code does not match your intention. Either use some other parameter to decide to add a Dense layer or add another argument to DenseNet to allow this behavior. 

Logically, it is counter intuitive to use a dense layer which obviously doesn't pool when the pool argument is not None and clearly states some form of pooling should occur. Such confusion gives rise to wrong models being built.

keras_contrib/applications/densenet.py
```diff
@@ -552,7 +592,8 @@ def __create_fcn_dense_net(nb_classes, img_input, include_top, nb_dense_block=5,
         assert reduction <= 1.0 and reduction > 0.0, "reduction value must lie between 0.0 and 1.0"
 
     # check if upsampling_conv has minimum number of filters
-    # minimum is set to 12, as at least 3 color channels are needed for correct upsampling
+    # minimum is set to 12, as at least 3 color channels are needed for
+    # correct upsampling
     assert nb_upsampling_conv > 12 and nb_upsampling_conv % 4 == 0, "Parameter `upsampling_conv` number of channels must " \
```
titu1994: I forgot to change this to >= 12. My mistake.

keras_contrib/applications/densenet.py
```diff
@@ -613,21 +656,26 @@ def __create_fcn_dense_net(nb_classes, img_input, include_top, nb_dense_block=5,
...
         else:
             out_shape[3] = n_filters_keep
 
         # upsampling block must upsample only the feature maps (concat_list[1:]),
-        # not the concatenation of the input with the feature maps (concat_list[0].
-        l = concatenate(concat_list[1:], axis=concat_axis)
+        # not the concatenation of the input with the feature maps
+        # (concat_list[0].
+        l = merge(concat_list[1:], mode='concat', concat_axis=concat_axis)
+        # l = concatenate(concat_list[1:], axis=concat_axis)
```
titu1994: Remove for now until the bug is fixed in Keras master

keras_contrib/applications/densenet.py
```diff
@@ -613,21 +656,26 @@ def __create_fcn_dense_net(nb_classes, img_input, include_top, nb_dense_block=5,
...
+        if K.image_data_format() == 'channels_first':
             out_shape[1] = n_filters_keep
         else:
             out_shape[3] = n_filters_keep
 
         # upsampling block must upsample only the feature maps (concat_list[1:]),
-        # not the concatenation of the input with the feature maps (concat_list[0].
-        l = concatenate(concat_list[1:], axis=concat_axis)
+        # not the concatenation of the input with the feature maps
+        # (concat_list[0].
```
titu1994: Remove comment  (concat_list[0]). should be after feature maps, not on a separate line

keras_contrib/applications/densenet.py
```diff
@@ -613,21 +656,26 @@ def __create_fcn_dense_net(nb_classes, img_input, include_top, nb_dense_block=5,
...
+        l = merge(concat_list[1:], mode='concat', concat_axis=concat_axis)
+        # l = concatenate(concat_list[1:], axis=concat_axis)
 
-        t = __transition_up_block(l, nb_filters=n_filters_keep, type=upsampling_type, output_shape=out_shape)
+        t = __transition_up_block(
```
titu1994: Should be on same line. No reason for new line for a small parameter list.

keras_contrib/applications/densenet.py
```diff
@@ -613,21 +656,26 @@ def __create_fcn_dense_net(nb_classes, img_input, include_top, nb_dense_block=5,
...
+            l, nb_filters=n_filters_keep, type=upsampling_type, output_shape=out_shape)
 
         # concatenate the skip connection with the transition block
-        x = concatenate([t, skip_list[block_idx]], axis=concat_axis)
+        x = merge([t, skip_list[block_idx]],
+                  mode='concat', concat_axis=concat_axis)
+        # x = concatenate([t, skip_list[block_idx]], axis=concat_axis)
```
titu1994: Remove concatenate code until bug is fixed in Keras master

titu1994:
@ahundt Has this been trained on the Pascal VOC dataset yet? My laptop is currently training another model ensemble for a class project, and estimated time is another 12-14 hours, so I can't train it myself.

As to the changes, they are minor and everything mostly looks good. One thing I am worried about is allowing depth parameter to be None. Could you explain how the code handle depth=None with non atrous pooling? It seems like you are overloading the meaning of depth parameter here.

keras_contrib/applications/densenet.py
ahundt: breaks pep8 length limit of 80 characters, required by upstream keras but not yet enabled for keras-contrib (same for other multi-line comments)

titu1994: Hmm, it looks absurd. How about putting dropout_rate to the top and weight_decay bottom?

keras_contrib/applications/densenet.py
ahundt: An alternative way to specify the depth is by the number of dense blocks and the number of layers per dense block. If those are both specified, then depth should be none.

Documented here:
https://github.com/farizrahman4u/keras-contrib/pull/46/files/9665d721cf38e6f01b3cb4cf7bc888f14230d9f4#diff-7c92da79e0a644cc9e206dab6de13babR55

keras_contrib/applications/densenet.py
ahundt: The goal is to support atrous convolutions, which will be a full image sized segmentation output (or a reduced one). There is no pooling in this case, and that is why the pooling option was added. In place of pooling there are atrous dilations to expand the receptive field, as if a 2x2 convolution covers the space of a 4x4 convolution when dilation_rate = 2.

I'll try to enumerate the important use cases:

 1. `pooling='max' || 'avg' && include_top=True || False` the final pooling layer is included, which is correct (look 2 lines above), perhaps pooling should be excluded when include_top is False? If this means moving line 530 and 531 to under line 533, that would seem reasonable. I was debating this before.
 2. `pooling=None && include_top=True` In this case the output will be 2D, even if dilation_rate=1, so the second 2D `top` is most appropriate, and GlobalAveragePooling would reduce the spatial resolution, which is undesirable for segmentation. 
 3. `pooling=None && include_top=False` Another 2D output case, pooling and top should both be entirely disabled, in this case the user will specify exactly what they want to do with the data, because it might be part of a completely different larger network.

keras_contrib/applications/densenet.py
ahundt: fixing it now

keras_contrib/applications/densenet.py
ahundt: Made this muli-line for pep8 compliance, I'll improve it a bit in the next commit

keras_contrib/applications/densenet.py
titu1994: Ok but add the assertion that i've mentioned below in the comment to ensure that at least 1 argument is properly given.

keras_contrib/applications/densenet.py
ahundt: Will be ready after https://github.com/fchollet/keras/pull/6035 is merged, which is likely quite soon. Also relevant https://github.com/fchollet/keras/pull/6034 and https://github.com/fchollet/keras/issues/5972. I might just wait a day or two instead of removing the comment. :-)

keras_contrib/applications/densenet.py
titu1994: The change you suggest makes sense. I would actually argue that a proper description of all 4 cases be included in the documentation of DenseNetFCN as examples, because I feel users will get confused between the interaction of pooling and include top. 

In the final places, pooling kind of refers to GlobalAveragePooling, whereas in earlier stages it refers to intermediate pooling steps (MaxPooling or AveragePooling). A good example documentation will go a long way in removing the confusion.

keras_contrib/applications/densenet.py
titu1994: Thanks.

keras_contrib/applications/densenet.py
titu1994: Thats alright. I would rather keep densenet code compatible with both, but for now I guess my only recourse is to keep a stable copy in my seperate repository so that I can continue using the code with Keras 1.2.2

ahundt:
@titu1994 Okay I realized I made a mistake in the transitions and the kernel_size was also needed for transitions if the dilation_rate were to be useful. I've added that plus the activation information, and reparameterized how top is specified, now include_top is specified separately from the kind of top. Hopefully now all is well and good!

ahundt:
@titu1994 I've made changes that integrate the functionality from your pull request, correctly parameterize the densenet transition layers, and should now make densenet.py meet the pep8 requirements so the build will no longer fail soon.

ahundt:
This is very out of date and the densenet + tensorflow memory issue detailed at https://github.com/tensorflow/tensorflow/issues/12948 hobbles this tremendously, so it isn't really practical to use. For those reasons I'm closing this, and if the situation changes a new PR with equivalent functionality can be created.

PR #52, opened 2017-03-27
[Bugfix] Fix assertion error when using DenseNetFCN with grayscale input + Add activation parameter

titu1994:
Fixes a assertion error when gray scale image is input to DenseNetFCN. 

The submitted code is basically a copy of the _check_input code in Keras, with the removal of a few parts which cause the crash.

### DIFF
--- a/keras_contrib/applications/densenet.py
+++ b/keras_contrib/applications/densenet.py
@@ -43,7 +43,7 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_filter=16, nb_layers
         optionally loading weights pre-trained
         on CIFAR-10. Note that when using TensorFlow,
         for best performance you should set
-        `image_dim_ordering="tf"` in your Keras config
+        `image_data_format='channels_last'` in your Keras config
         at ~/.keras/keras.json.
 
         The model and the weights are compatible with both
@@ -102,7 +102,7 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_filter=16, nb_layers
     input_shape = _obtain_input_shape(input_shape,
                                       default_size=32,
                                       min_size=8,
-                                      data_format=K.image_dim_ordering(),
+                                      data_format=K.image_data_format(),
                                       include_top=include_top)
 
     if input_tensor is None:
@@ -132,13 +132,13 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_filter=16, nb_layers
                 (bottleneck is False) and (reduction == 0.0) and (dropout_rate == 0.0) and (weight_decay == 1E-4):
             # Default parameters match. Weights for this model exist:
 
-            if K.image_dim_ordering() == 'th':
+            if K.image_data_format() == 'channels_first':
                 if include_top:
-                    weights_path = get_file('densenet_40_12_th_dim_ordering_th_kernels.h5',
+                    weights_path = get_file('densenet_40_12_th_data_format_th_kernels.h5',
                                             TH_WEIGHTS_PATH,
                                             cache_subdir='models')
                 else:
-                    weights_path = get_file('densenet_40_12_th_dim_ordering_th_kernels_no_top.h5',
+                    weights_path = get_file('densenet_40_12_th_data_format_th_kernels_no_top.h5',
                                             TH_WEIGHTS_PATH_NO_TOP,
                                             cache_subdir='models')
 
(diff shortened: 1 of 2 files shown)


titu1994:
The tests are failing due to a "E303 too many blank lines (2)" in the PEP test which I have fixed. The other errors are due to Keras 2 code errors in test_deconvolution_3d and test_advanced_activations

titu1994:
I've added an activation parameter, which allows use of sigmoid activation when training a network on binary segmentation masks. This is more suitable than a class wise loss over individual pixels.

keras_contrib/applications/densenet.py
```diff
@@ -132,13 +132,13 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_filter=16, nb_layers
                 (bottleneck is False) and (reduction == 0.0) and (dropout_rate == 0.0) and (weight_decay == 1E-4):
             # Default parameters match. Weights for this model exist:
 
-            if K.image_dim_ordering() == 'th':
+            if K.image_data_format() == 'channels_first':
                 if include_top:
-                    weights_path = get_file('densenet_40_12_th_dim_ordering_th_kernels.h5',
+                    weights_path = get_file('densenet_40_12_th_data_format_th_kernels.h5',
```
ahundt: did you upload files to the new path? Same applies to other weight files

keras_contrib/applications/densenet.py
```diff
@@ -250,11 +252,18 @@ def DenseNetFCN(input_shape, nb_dense_block=5, growth_rate=16, nb_layers_per_blo
...
+        raise ValueError('Parameter "activation" must be one of "softmax" or "sigmoid"')
+
+    if activation == 'sigmoid' and classes != 1:
+        raise ValueError('Parameter "activation" can be "sigmoid" only when parameter "classes" is equal to 1')
+
     # Determine proper input shape
+    min_size = 2 ** nb_dense_block
     input_shape = _obtain_input_shape(input_shape,
                                       default_size=32,
                                       min_size=16,
```
ahundt: should min_size be passed here?

keras_contrib/applications/densenet.py
```diff
@@ -43,7 +43,7 @@ def DenseNet(depth=40, nb_dense_block=3, growth_rate=12, nb_filter=16, nb_layers
         optionally loading weights pre-trained
```
ahundt: Add activation parameter to DenseNet as well. Not only is it useful in regular DenseNet for the two class case, it will also be very useful when combined with Atrous DenseNet in #46.

keras_contrib/applications/densenet.py
```diff
@@ -406,7 +415,7 @@ def __transition_up_block(ip, nb_filters, type='upsampling', output_shape=None,
     elif type == 'subpixel':
         x = Conv2D(nb_filters, (3, 3), padding="same", kernel_regularizer=l2(weight_decay), activation='relu',
                    use_bias=False, kernel_initializer='he_uniform')(ip)
-        #x = Convolution2D(nb_filters, 3, 3, activation="relu", border_mode='same', W_regularizer=l2(weight_decay),
```
ahundt: These comments shouldn't be here, and all calls should use the updated Keras-2 API... You might just want to make use of the similar changes in #46

keras_contrib/applications/densenet.py
titu1994: No I did not upload any new file to a new path. The path remains the same, just the name of the weight file changes when it is downloaded onto a local machine.

keras_contrib/applications/densenet.py
titu1994: Seems the code did not update. Will fix in next commit.

keras_contrib/applications/densenet.py
titu1994: I'm keeping Keras 1 code for now, since it is cross compatible with Keras 2. Not the other way around.

I need to use Keras 1.2.2 for the time being, since I need many features which were abruptly removed for random reasons from Keras 2 (such as Batch norm mode 2 which I need for GANs, fbeta_score metrics, and a few bugs which have come up in 2 are making me wait for the shift). 

I will update the code when Keras 2 stabilizes (bug free), and fixes some of its mistakes like Mode 2 batchnorm. Until then, better to keep code compatible with both versions.

titu1994:
Someone merged the Keras 2 api code changes to this, and so I will rewrite this PR. Also, who merged the changes related to concatenate when the master Keras is showing bugs with concatenate in DenseNet??

PR #53, opened 2017-03-28
Fit DenseNet examples to Keras-2

kemaswill:


ahundt:
https://github.com/fchollet/keras/pull/6034 may be relevant. Also a much more complete version of these changes is in https://github.com/farizrahman4u/keras-contrib/pull/46/files#diff-7c92da79e0a644cc9e206dab6de13bab

farizrahman4u:
LGTM, thanks!

keras_contrib/applications/densenet.py
```diff
@@ -375,15 +376,16 @@ def __dense_block(x, nb_layers, nb_filter, growth_rate, bottleneck=False, dropou
         x = __conv_block(x, growth_rate, bottleneck, dropout_rate, weight_decay)
         x_list.append(x)
 
-        x = merge(x_list, mode='concat', concat_axis=concat_axis)
+        x1 = concatenate(x_list, axis=concat_axis)
```
ahundt: Not good! then all the convolutions will be the same! it is essential that all the concatenation be on x.

ahundt: Specifically, this is the source of fchollet/keras#6034

kemaswill: Hmm, interesting. So I should wait for this PR to be merged, right?

ahundt: Unfortunately, this change both triggers an existing bug and adds a new one. Consider just utilizing/merging #46 it fixes everything mentioned here, and I'll be switching to concat calls once fchollet/keras#6034 is merged.

kemaswill: Sure, I will take a look. Thanks.

kemaswill:
@ahundt Has you code be merged? If yes, then we don't need the revert any more, right?

ahundt:
@kemaswill I'm still waiting on review for #46

kemaswill:
Got it.

PR #55, opened 2017-03-29
[Bugfix] Corrects activation for DenseNetFCNs

titu1994:
Does 3 things:
- Fixes a bug in DenseNetFCN where classes = 1 and softmax cause learning to fail
- Adds activation parameter to both DenseNet and DenseNetFCN to allow training on grayscale images


titu1994:
@ahundt This PR includes the changes you requested, could you review it again?

ahundt:
LGTM

tboquet:
Should this also be the most PEP8 compliant possible?
The application is really cool but has a lot of lines. It will ease the reading.

titu1994:
This is pep 8 compliant. It passes the pep 8 test in my local machine, but you need to set pep to ignore line too long error for the 4 weight file paths. Other than that, all other parts of the code are compliant.

ahundt:
@titu1994 perhaps just copy the pep8ified link lines from #46

farizrahman4u:
Why are we reverting to Keras-1? The concat bug has been fixed.

titu1994:
@farizrahman4u You're right. This was a pr for when the bug hadn't been fixed. I'll push the changes to the new api in some time.

titu1994:
@farizrahman4u Pushed the update. It now uses Keras 2 api. It is ready to merge

farizrahman4u:
It would be great if one more maintainer could review the new changes.

titu1994:
Yes that's a good idea. I tested just the model building code and the tests for the layers and backend, but the code itself could go through one more review.

keras_contrib/applications/densenet.py
```diff
@@ -376,56 +385,46 @@ def __dense_block(x, nb_layers, nb_filter, growth_rate, bottleneck=False, dropou
...
     if type == 'upsampling':
         x = UpSampling2D()(ip)
     elif type == 'subpixel':
-        x = Conv2D(nb_filters, (3, 3), padding="same", kernel_regularizer=l2(weight_decay), activation='relu',
+        x = Conv2D(nb_filters, (3, 3), activation="relu", padding='same', W_regularizer=l2(weight_decay),
                    use_bias=False, kernel_initializer='he_uniform')(ip)
-        #x = Convolution2D(nb_filters, 3, 3, activation="relu", border_mode='same', W_regularizer=l2(weight_decay),
-        #                  bias=False, init='he_uniform')(ip)
         x = SubPixelUpscaling(scale_factor=2)(x)
-        x = Conv2D(nb_filters, (3, 3), activation="relu", padding='same', kernel_regularizer=l2(weight_decay),
```
ahundt: Is the removal of this second convolution in the subpixelupscaling case correct?

Also does the subpixel upscaling even work & is it necessary? The 100 layers paper describes its dense blocks after each upscaling step as essentially providing this functionality.

keras_contrib/applications/densenet.py
```diff
@@ -604,36 +600,19 @@ def __create_fcn_dense_net(nb_classes, img_input, include_top, nb_dense_block=5,
 
     skip_list = skip_list[::-1]  # reverse the skip list
 
-    if K.image_dim_ordering() == 'th':
```
ahundt: what makes out_shape no longer necessary again?

keras_contrib/applications/densenet.py
```diff
@@ -245,17 +244,33 @@ def DenseNetFCN(input_shape, nb_dense_block=5, growth_rate=16, nb_layers_per_blo
...
 
     # Determine proper input shape
-    input_shape = _obtain_input_shape(input_shape,
-                                      default_size=32,
-                                      min_size=16,
-                                      data_format=K.image_dim_ordering(),
-                                      include_top=include_top)
+    min_size = 2 ** nb_dense_block
+
+    if K.image_dim_ordering() == 'th':
```
ahundt: should be `'channels_first'` or `'channels last'` in keras-2, this applies to all other instances of `tf` and `th` in these changes.

keras_contrib/applications/densenet.py
titu1994: The second convolutional is below before the else. 

Also, the dense block is not the cause of increase in spatial dimension. They use Deconvolution to achieve that. However, simple upscaling layer of Keras is useful to about border checkerboard effects and subpixel is better for large images and also avoids checkerboard effects

keras_contrib/applications/densenet.py
titu1994: Keras 2 deciphers the output shape of Deconvolution layers automatically​ now. Thus it is no longer to compute the output shape for which we needed a static batch size (none batch size did not work with Tensorflow before )

keras_contrib/applications/densenet.py
titu1994: Yeah my bad. Will fix once in some time.

titu1994:
@ahundt Changed all image_dim_ordering to image_data_format.

keras_contrib/applications/densenet.py
```diff
@@ -135,7 +135,7 @@ def DenseNet(input_shape=None, depth=40, nb_dense_block=3, growth_rate=12, nb_fi
                 (bottleneck is False) and (reduction == 0.0) and (dropout_rate == 0.0) and (weight_decay == 1E-4):
             # Default parameters match. Weights for this model exist:
 
-            if K.image_dim_ordering() == 'th':
```
ahundt: not correct, 'th' should be 'channels_first', same everywhere else

ahundt: see #46 for actual correct version of this change

titu1994: Fixed. Sorry about that.

keras_contrib/applications/densenet.py
```diff
@@ -537,9 +537,9 @@ def __create_fcn_dense_net(nb_classes, img_input, include_top, nb_dense_block=5,
     Returns: keras tensor with nb_layers of conv_block appended
     '''
 
-    concat_axis = 1 if K.image_data_format() == "th" else -1
+    concat_axis = 1 if K.image_data_format() == "channels_first" else -1
```
ahundt: quotes should always be single (this is what the rest of keras does)

ahundt:
@titu1994 I think with a quick merge of master, which would mean the build should pass, and substituting the double quotes for single quotes this could be merged.

titu1994:
Sorry I'm a bit busy with assignments and tests. Will fix those in the coming days.

ahundt:
no worries I understand how that goes

titu1994:
The errors in the build are due to some fault in Keras K.in_batch_training implementation in Keras 2 (due to my older BatchRenormalization code)

Will create a seperate PR to upgrade Batch Renormalization to Keras 2 in a few days.

titu1994:
All tests passing now since Batch Renormalization has been merged. This PR is ready to be merged now. 
@farizrahman4u @the-moliver

PR #57, opened 2017-03-29
Fix: Namings in test_convolutional.py

t-ae:
All appearances of `b_` are replaced with `bias_` in [this commit](https://github.com/farizrahman4u/keras-contrib/commit/3773baa1fba7b52649531c001ee488a28421e424).
Fixed them.


farizrahman4u:
LGTM, Thanks.

PR #58, opened 2017-03-29
Improve `depth_to_space`

t-ae:
Added test of `depth_to_space`.

Then I found Theano's implementation didn't pass the test.
(TF uses its own `depth_to_space`, so test itself is correct, see also [this discussion](https://github.com/farizrahman4u/keras-contrib/pull/34/files/6c656f12d967901e0cb9965f9cf783f280631312#diff-091b2f7e9b1a041322f576cf6e448153))
I replaced it with the simple implementation.
(I'm not confident it's also better or not in performance...)

### DIFF
--- a/keras_contrib/backend/theano_backend.py
+++ b/keras_contrib/backend/theano_backend.py
@@ -190,13 +190,10 @@ def depth_to_space(input, scale, data_format=None):
     input = _preprocess_conv2d_input(input, data_format)
 
     b, k, row, col = input.shape
-    output_shape = (b, k // (scale ** 2), row * scale, col * scale)
-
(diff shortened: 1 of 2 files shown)


farizrahman4u:
LGTM. Can someone else review too?

ahundt:
LGTM, also compared to the [tensorflow batch_to_space test](https://github.com/tensorflow/tensorflow/blob/584d4921014db921a8f4722749adff09737826a8/tensorflow/python/kernel_tests/batchtospace_op_test.py#L49) for reference.

the-moliver:
LGTM too!

PR #60, opened 2017-03-30
Implemented convolution aware initialization

ArmenAg:
Hello,

I implemented convolution aware initialization as described in this paper: https://arxiv.org/abs/1702.06295

Thank you.

### DIFF
--- a/keras_contrib/initializers.py
+++ b/keras_contrib/initializers.py
@@ -56,7 +56,7 @@ class ConvolutionAware(Initializer):
             correct_fft = np.fft.rfftn
             correct_ifft = np.fft.irfftn
         else:
-            return self.orthogonal(shape)
+            return K.variable(self.orthogonal(shape), dtype=K.floatx())
 
         kernel_fourier_shape = correct_fft(np.zeros(kernel_shape)).shape
 
@@ -73,7 +73,7 @@ class ConvolutionAware(Initializer):
             init.append(filters)
 
         # Format of array is now: filters, stack, row, column
-        init = np.array(init, dtype=K.floatx())
+        init = np.array(init)
         init = self._scale_filters(init, variance)
         return init.transpose(transpose_dimensions)
 
--- a/keras_contrib/initializers.py
+++ b/keras_contrib/initializers.py
@@ -2,3 +2,95 @@ from __future__ import absolute_import
 import numpy as np
 from . import backend as K
 from keras.initializers import *
+from keras.initializers import _compute_fans
+
+
+class ConvolutionAware(Initializer):
+    """
+    Initializer that generates orthogonal convolution filters in the fourier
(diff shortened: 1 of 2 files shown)


the-moliver:
Looks good, but is there any reason not to extend to 3D filters as well? Seems easy enough to do...

ArmenAg:
In the paper I really only focused on 1D and 2D convolution, so there's not empirical evidence that it would work for 3D convolution (although in theory it should). I could implement it using numpy's irfftn and rfftn functions. Would you recommend doing so?

the-moliver:
Because it's easy to do and potentially useful, I would recommend it.

ArmenAg:
I just added support for Conv3D. Let me know what you think.

ArmenAg:
I added tests that reflected the upper and lower bounds proved in the paper. I had to add a new parameters to the _runner to allow testing against bounds.

the-moliver:
LGTM!

ArmenAg:
Great! When will it be merged?

the-moliver:
Very soon! Was hoping to get another review, just sent out requests since it hadn't happened spontaneously.

iperov:
if someone want
my super fast implementation of convolution aware for Conv2D (4d shapes)

```
def generate(shape, eps_std=0.05, dtype=np.float32):
    """
    Super fast implementation of Convolution Aware Initialization for 4D shapes
    Convolution Aware Initialization https://arxiv.org/abs/1702.06295
    """
    if len(shape) != 4:
        raise ValueError("only shape with rank 4 supported.")
    row, column, stack_size, filters_size = shape
    fan_in = stack_size * (row * column)
    kernel_shape = (row, column)
    kernel_fft_shape = np.fft.rfft2(np.zeros(kernel_shape)).shape
    basis_size = np.prod(kernel_fft_shape)
    if basis_size == 1:
        x = np.random.normal( 0.0, eps_std, (filters_size, stack_size, basis_size) )
    else:
        nbb = stack_size // basis_size + 1
        x = np.random.normal(0.0, 1.0, (filters_size, nbb, basis_size, basis_size))
        x = x + np.transpose(x, (0,1,3,2) ) * (1-np.eye(basis_size))
        u, _, v = np.linalg.svd(x)
        x = np.transpose(u, (0,1,3,2) )
        x = np.reshape(x, (filters_size, -1, basis_size) )
        x = x[:,:stack_size,:]
    x = np.reshape(x, ( (filters_size,stack_size,) + kernel_fft_shape ) )
    x = np.fft.irfft2( x, kernel_shape ) \
        + np.random.normal(0, eps_std, (filters_size,stack_size,)+kernel_shape)
    x = x * np.sqrt( (2/fan_in) / np.var(x) )
    x = np.transpose( x, (2, 3, 1, 0) )
    return x.astype(dtype)
```

PR #66, opened 2017-04-10
[WIP] Add LSUN dataset

souravsingh:
Work in Progress for adding the LSUN dataset for classification and segmentation, which is accessible [here](http://www.yf.io/p/lsun).


nzw0301:
[`get_file`](https://keras.io/utils/#get_file) may be a good way to download file for unification.

PR #67, opened 2017-04-11
updated TF backend - redundant line in deconv3d causing errors

gattia:
I was getting errors about miss match between deconvolution kernel size and input image size. 

After looking at print outs of the image size and the kernel at different levels it was apparent that the kernel being inputted into `x = tf.nn.conv3d_transpose(x, kernel, output_shape, strides, padding=padding)` at line 117 in tensorflow_backend.py was wrong and that the image dimensions were correct. 

It appears as though the manipulation of the kernel `kernel = _preprocess_conv3d_kernel(kernel, data_format)` at line 112 was extraneous because either 1) was never needed or 2) there is now a manipulation in convolution.py that makes it redundant. 

removing or commenting out line 112 makes it run without error.


the-moliver:
I'm going to merge this for now so we can see when new things pass tests, but still working on a refactor.

PR #69, opened 2017-04-15
Update BatchRenorm to Keras 2 API

titu1994:
Updates the BatchRenormalization layer to Keras 2 specifications and should fix the tests as well.

### DIFF
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -1,5 +1,5 @@
 from keras.engine import Layer, InputSpec
-from .. import initializers, regularizers
+from .. import initializers, regularizers, constraints
 from .. import backend as K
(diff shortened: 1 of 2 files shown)


titu1994:
Theres some bug with Theano. It calls something like ifelse.ifelse which doesnt exist. Is this repository using the master Theano? Keras proper doesnt have errors like this.

the-moliver:
It seems Keras uses the Theano release for tests while we are on bleeding edge here. This seems likely to be the problem. Please submit a bug report to Theano so they can fix. Also I made a PR to switch to using Theano release for tests: https://github.com/farizrahman4u/keras-contrib/pull/72

the-moliver:
Yup that fixed it.

titu1994:
Cool. Thanks.

PR #74, opened 2017-04-18
Fix keras argument spelling error in densenet.py

DNGros:
It looks like in 1d57dd3c63eab5b181c412436c9fcffc748cc72e a spelling error ("kenel" not "kernal") was introduced into densenet.py that causes an argument error from keras when trying to use the network. This pull request corrects that.


ahundt:
your spelling correction isn't quite correct. I believe `kernel` is the correct spelling.

ahundt:
@the-moliver the code is still broken, I don't believe this fix which was merged works.

PR #75, opened 2017-04-19
Adds improved_wgan.py to examples

Rocketknight1:
improved_wgan.py implements the improved WGAN from https://arxiv.org/abs/1704.00028 in Keras. Since Keras is frequently used by new learners of neural networks, and since gradient penalties are an unusual thing to implement, I've endeavoured to heavily document the code with extended docstrings and comments so that people can learn from it.


farizrahman4u:
Haven't gone through the code, but the documentation looks neat. Can anyone please review?

leriomaggio:
Apparently the PEP8 check failed! Try using some pep8 checkers to resolve those issues and it should be ok!

For instance, one I caught at glance: `Random_Weighted_Average` --> `RandomWeightedAverage`

HTH

Rocketknight1:
I blame PyCharm not telling me about trailing whitespace. All fixed!

tboquet:
Nice example and implementation!

the-moliver:
Looks good. Can you @Rocketknight1 (or someone) please test that it works with Theano. It should, since you only use Keras backend functions, but I'd be nice to have confirmation before merging.

Rocketknight1:
The only unusual thing in the network is the use of K.gradients() in the gradient penalty loss function, which means that calculating that loss requires second-order gradients. I'm not sure how it'll work in Theano, but I'll try it.

Rocketknight1:
I got it running with Theano (CPU only). Compiled and ran fine, though I wasn't able to run it long enough to see any real training progress.

examples/improved_wgan.py
```diff
@@ -0,0 +1,263 @@
...
+    # are then run through the discriminator. Although we could concatenate the real and generated images into a
+    # single tensor, we don't (see model compilation for why).
+    real_samples = Input(shape=X_train.shape[1:])
+    generator_input_for_discriminator = Input(shape=(100,))
+    generated_samples_for_discriminator = generator(generator_input_for_discriminator)
+    discriminator_output_from_generator = discriminator(generated_samples_for_discriminator)
+    discriminator_output_from_real_samples = discriminator(real_samples)
+
+    # We also need to generate weighted-averages of real and generated samples, to use for the gradient norm penalty.
+    averaged_samples = RandomWeightedAverage()([real_samples, generated_samples_for_discriminator])
```
the-moliver: Can you explain what you are doing here and why? I don't see anything like this in the paper. It seems like you are averaging the pixels of generated samples and real samples and then feeding to the discriminator. This doesn't seem right at all.

examples/improved_wgan.py
```diff
@@ -0,0 +1,263 @@
...
+    model.add(BatchNormalization())
+    model.add(LeakyReLU())
+    model.add(Conv2DTranspose(64, (5, 5), strides=2, padding='same'))
+    model.add(BatchNormalization())
+    model.add(LeakyReLU())
+    # Because we normalized training inputs to lie in the range [-1, 1],
+    # the tanh function should be used for the output of the generator to ensure its output
+    # also lies in this range.
+    model.add(Convolution2D(1, (5, 5), padding='same', activation='tanh'))
+    model.add(Reshape((28, 28, 1)))
```
the-moliver: Please make this respect the `image_data_format`

examples/improved_wgan.py
```diff
@@ -0,0 +1,263 @@
...
+
+def make_discriminator():
+    """Creates a discriminator model that takes an image as input and outputs a single value, representing whether
+    the input is real or generated. Unlike normal GANs, the output is not sigmoid and does not represent a probability!
+    Instead, the output should be as large and negative as possible for generated inputs and as large and positive
+    as possible for real inputs.
+
+    Note that the improved WGAN paper suggests that BatchNormalization should not be used in the discriminator."""
+    model = Sequential()
+    model.add(Convolution2D(64, (5, 5), padding='same', input_shape=(28, 28, 1)))
```
the-moliver: Also here please respect the `image_data_format` choice

Rocketknight1: Apologies, I have to stop assuming that everyone's working in Tensorflow style. I'll update that soon.

examples/improved_wgan.py
Rocketknight1: This is definitely the most confusing part of the code, but I'm fairly sure I've implemented it correctly. You are correct that I'm averaging real and generated samples, though.

The core idea is that you want the gradient norm to be close to 1 everywhere; this is the essence of the Lipschitz constraint that must be met for WGANs to train properly. However, it is impossible to guarantee that the gradient norm for every possible input will be close to 1. The compromise used in the paper is that you "Sample along straight lines" between real and generated samples. Equations (4) and (5) on page 6 describe how they do this, which you will see is equivalent to the randomly-weighted average between real and generated samples from my function. It is effectively equivalent to plotting a line between a real sample and a generated sample, and sampling a random point on that line with uniform probability.

The reason the randomly-averaged functions are passed through the discriminator afterwards is not because we particularly care about their output. Instead, we care about the _gradient_ we get, because we want to ensure its norm stays close to 1. When the model is compiled, you'll notice the loss for the randomly-averaged samples is not the wasserstein_loss, but the partial_gp_loss function.

examples/improved_wgan.py
```diff
@@ -0,0 +1,264 @@
...
+    test_image_stack = generator_model.predict(np.random.rand(10, 100))
+    test_image_stack = (test_image_stack * 127.5) + 127.5
+    test_image_stack = np.round(test_image_stack).astype(np.uint8).reshape(test_image_stack.shape[:-1])
+    tiled_output = tile_images(test_image_stack)
+    tiled_output = Image.fromarray(tiled_output, mode='L')  # L specifies greyscale
+    outfile = os.path.join(output_dir, 'epoch_{}.png'.format(epoch))
+    tiled_output.save(outfile)
+
+
+def main():
```
Rocketknight1: Fixed, thank you!

examples/improved_wgan.py
Rocketknight1: Done!

examples/improved_wgan.py
Rocketknight1: Done!

examples/improved_wgan.py
```diff
@@ -0,0 +1,265 @@
...
+
+    In order to evaluate the gradients, we must first run samples through the discriminator and evaluate the loss.
+    Then we get the gradients of the trainable discriminator weights, flatten the gradient tensors
+    and concatenate them into one 1D gradient tensor. The l2 norm and penalty can then be calculated from it.
+
+    Note that this loss function requires the model's trainable weights as input, but Keras only supports passing
+    y_true and y_pred to loss functions. To get around this, we make a partial() of the function with the
+    trainable_weights argument, and use that for model training."""
+    averaged_samples_loss = K.mean(y_pred)
+    gradients = K.gradients(averaged_samples_loss, trainable_weights)
```
the-moliver: After reading the paper more closely, I believe the penalty should be applied to the gradient of the discriminator with respect to the input to the discriminator, ie your averaged samples. From the top of page 6: "The gradient term ||∇xˆD(xˆ)||2 is with respect to the points xˆ, not the parameters of D." I still find this averaging of samples to be quite a weird idea :-)

Rocketknight1: Haha, yeah, it's definitely the first time I've come across that in the deep learning literature. It's part of what made this such an interesting model to implement.

Also, you might be right! I was getting very good empirical results with the existing code, but looking at the maths again, I did indeed overlook this. I'll try replacing that line with K.gradients(averaged_samples_loss, averaged_samples) and see how it performs while I double check things.

Rocketknight1: I double-checked your workings and you're right, I was taking the gradient of the discriminator w.r.t. the weights, not the inputs. Lesson learned there, I guess: Just because empirical performance is good doesn't mean you haven't left a nasty maths error somewhere in the code. Performance with the change seems roughly similar, possibly slightly faster convergence?

the-moliver: That is indeed an important lesson. I've been surprised how well things can work with even bigger errors than that!

examples/improved_wgan.py
```diff
@@ -72,7 +73,7 @@ def gradient_penalty_loss(y_true, y_pred, trainable_weights):
     y_true and y_pred to loss functions. To get around this, we make a partial() of the function with the
     trainable_weights argument, and use that for model training."""
     averaged_samples_loss = K.mean(y_pred)
-    gradients = K.gradients(averaged_samples_loss, trainable_weights)
+    gradients = K.gradients(averaged_samples_loss, averaged_samples)
```
the-moliver: I think this should be `gradients = K.gradients(y_pred, averaged_samples)` "since we penalize the norm of
the critic’s gradient with respect to each input independently, and not the entire batch". In other words you're calculating the gradient of the output of the discriminator (not a loss computed on the output which averages across a batch).

Rocketknight1: You're right, though I think using K.mean() might just result in constant scaling of the gradient compared to the correct version? (Specifically scaling it down by BATCH_SIZE). Still, I'm not sure the mathematical world is ready for BATCH_SIZE-Lipschitz functions just yet, so I'll make the change :p

Rocketknight1:
@the-moliver One thing I noticed post-merging: I had been doubling the gradient penalty hyperparameter in the gradient_penalty loss function because the size of averaged_samples batch was half the size of the combined real and generated samples being passed to the discriminator. However, on double-checking the paper, I believe that they had already accounted for this in setting the penalty hyperparameter, because their pseudocode in the paper matches my final algorithm quite closely: The total loss is the Wasserstein loss on one batch of real samples and one batch of generated samples, plus gradient penalty loss on one batch of averaged samples, all of equal size.

Can you double-check to confirm that I'm correct, and if so I'll push a change with the correct gradient penalty hyperparam?

the-moliver:
Yeah I think you should get rid of the factor of 2 you have in there. Also it's probably a good idea to make `gradient_penalty_loss` take `GRADIENT_PENALTY_WEIGHT` as keyword argument rather than using a global variable.

PR #76, opened 2017-04-21
CRF for keras 2.x

linxihui:
An Implementation of linear chain conditional random field (CRF), modified for keras 2.0 from previous PR https://github.com/fchollet/keras/pull/4646

Note this PR has https://github.com/farizrahman4u/keras-contrib/pull/79 merged. Please merged.

### DIFF
--- a/examples/conll2000_chunking_crf.py
+++ b/examples/conll2000_chunking_crf.py
@@ -1,0 +1,99 @@
+"""Train CRF and BiLSTM-CRF on CONLL2000 chunking data, similar to https://arxiv.org/pdf/1508.01991v1.pdf.
+"""
+from __future__ import absolute_import
+from __future__ import print_function
+
+import numpy
+from collections import Counter
+from sklearn.metrics import classification_report
+from nltk.corpus import conll2000
+
+from keras.models import Sequential
+from keras.layers import Embedding, Bidirectional, LSTM
+from keras.preprocessing.sequence import pad_sequences
+from keras_contrib.layers import CRF
+
+EPOCHS = 10
+TRAIN_VALID_SPLIT = 0.8
+EMBED_DIM = 200
+BiRNN_UNITS = 200
+
+numpy.random.seed(1223)
+
+if not hasattr(conll2000, 'tagged_sents'):
+    import nltk
+    nltk.download('conll2000')
+
+tagged_sents = [s for s in conll2000.tagged_sents()]
+class_labels = sorted(list(set(w[1] for s in tagged_sents for w in s)))
+
+N = len(tagged_sents)
+train_nb = int(N * TRAIN_VALID_SPLIT)
+idx = numpy.random.choice(numpy.arange(N), N, replace=False)
+train = [tagged_sents[i] for i in idx[:train_nb]]
+test = [tagged_sents[i] for i in idx[train_nb:]]
+
+word_counts = Counter(w[0].lower() for s in train for w in s)
+vocab = ['<pad>', '<unk>'] + [w for w, f in word_counts.iteritems() if f >= 3]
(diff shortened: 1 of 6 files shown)


linxihui:
@farizrahman4u The tensorflow and keras version are not compatible in .travis checking.
The tensorflow version is 0.12. It has two concatenate functions `tf.concat(axis, tensors)` and `tf.concat_v2(tensors, axis)`. However, in keras tensorflow backend, `K.concatenate(tensors, axis) = tf.concat(tensors, axis)`. In tensorflow 1.x, `tf.concat` is changed to accept `tensors` as 1st argument.  This is the reason the CRF checking fail.

I think tensorflow should be update to version 1.x in .travis.

farizrahman4u:
@linxihui Thanks for pointing it out.. please submit a PR to fix travis config so that you get the Github credits.

farizrahman4u:
Cool stuff! Can anyone review?

keras_contrib/layers/crf.py
```diff
@@ -0,0 +1,542 @@
...
+        def acc(y_true, y_pred):
+            X = self.inbound_nodes[0].input_tensors[0]
+            mask = self.inbound_nodes[0].input_masks[0]
+            y_pred = self.get_marginal_prob(X, mask)
+            return self._get_accuracy(y_true, y_pred, mask, self.sparse_target)
+        acc.func_name = 'marginal_acc'
+        return acc
+
+    @staticmethod
+    def log_sum_exp(x, axis=-1):
```
tboquet: Since fchollet/keras#6346 has been merged, could it be integrated in this PR?

examples/conll2000_chunking_crf.py
```diff
@@ -0,0 +1,99 @@
+"""Train CRF and BiLSTM-CRF on CONLL2000 chunking data, similar to https://arxiv.org/pdf/1508.01991v1.pdf.
+"""
+from __future__ import absolute_import
+from __future__ import print_function
+
+import numpy
+from collections import Counter
+from sklearn.metrics import classification_report
```
tboquet: Should we try to avoid those dependencies (`sklearn` and `nltk`)?
I agree this is very convenient and not hard to install but it could be easy to add a new dataset as in #25.

keras_contrib/layers/crf.py
linxihui: Sure, will change it soon.

examples/conll2000_chunking_crf.py
linxihui: `sklearn` is not necessary, as I just use the `classification_report` to print the result in a nice way. `nltk` is here for getting the data. My  original idea was a bit on the opposite, avoiding having the same dataset in multiple packages or writing duplicate script to load data, but instead using an existing package, which also saves the effort of licensing.  But I understand the disadvantage as well. I'll change it, as it probably fits better to the Keras way.

keras_contrib/layers/crf.py
tboquet: Ok let's keep this in mind and open an issue when this will be merged?

linxihui: Then I probably should just change it.  Actually, there is a bug when using `K.logsumexp` with` `tensorflow` backend, when passed a placeholder, it losses the shape information, i.e., `K.int_shape` got None. This is due to the fact that `K.logsumexp` uses an deprecidated argument. I'll make a PR to `keras` and will make the change here one it is merged.

linxihui:
@tboquet  using `K.logsumexp` now.

linxihui:
@tboquet  @farizrahman4u  any one do the code review?

farizrahman4u:
LGTM, need confirmation from one more moderator.

PR #78, opened 2017-04-25
Correct loss weight hyperparam for improved_wgan

Rocketknight1:


Rocketknight1:
@the-moliver I've added the gradient penalty fix; though I think I could probably have made a more elegant pull request for it.

the-moliver:
Changes incorporated!

Issue #47, opened 2017-03-17
additional datasets - coco + voc2012


ahundt:
I'm working on some code that loads ms coco and pascal voc for segmentation, but it would currently involve steps that depend on several other python libraries, including tf only steps. It could be used to create examples for the DenseNetFCN and #46 segmentation models. 

I'm creating an example that trains on these keras-contrib models but with the dependency limitations mentioned above, and I can modify it for the keras-contrib code structure.  However, this brings up the following questions:

 - Would keras-contrib accept a contribution with those limitations?
 - What would be the procedure?
 - Is someone interested in adapting for more general operation?

If it makes the most sense for me to just keep everything separate, that's okay.

patyork:
What are the external libraries (and what are they used for)? Also, what are the steps that can only be done in TF?

ahundt:
I'm sure they can be done with other tools, they are currently implemented with TF and would need to be rewritten to work with numpy arrays. Actually it probably could be done fairly easily without TF I've been using it a bit more since the previous post.

Pascal VOC 2012 + Berkeley additional labeling

- [pascal_voc.py](https://github.com/ahundt/tf-image-segmentation/blob/tf_densenet_fcn/tf_image_segmentation/utils/pascal_voc.py)
- skimage for loading matlab matrices and converting to png


Extra for [coco](mscoco.org):
 - [pycocotools](https://github.com/pdollar/coco/blob/master/PythonAPI/pycocotools/coco.py) which

ahundt:
I've made useful progress towards Pascal VOC and MS COCO in other repositories, which I plan to merge into keras-contrib when ready. Details are in https://github.com/ahundt/Keras-FCN/tree/densenet_atrous

One issue is that with more complex datasets it makes sense to have several manual utilities & external dependencies may be required. 

How can the manual utility + dependency needs of large datasets be reconciled with the need to avoid dependencies when possible with Keras? 

How should a single dataset dependency, like [pycocotools](https://github.com/pdollar/coco/blob/master/PythonAPI/pycocotools/coco.py), be handled?

ahundt:
#80 adds pascal_voc support

ahundt:
#81 adds ms coco support

ahundt:
Some improvments to #81 should be made based on https://github.com/PavlosMelissinos/enet-keras/blob/master/src/data/datasets.py

tarvaina:
[...]

PR #81, opened 2017-04-26
coco.py mscoco dataset support

ahundt:
mscoco dataset support
#47 

This implementation works correctly, but some performance and space saving measures may be desirable, including directly using the pycocotools to create any masks so it can all simply be kept in memory, rather than the current procedure of writing to disk. Such improvements can likely be incorporated from https://github.com/PavlosMelissinos/enet-keras/blob/master/src/data/datasets.py

depends on [sacred](https://github.com/IDSIA/sacred/) for the command line utilities

additional direct/indirect contributors include @PavlosMelissinos, @vittalP, @warmspringwinds

### DIFF
--- a/keras_contrib/datasets/coco.py
+++ b/keras_contrib/datasets/coco.py
@@ -1,0 +1,440 @@
+#!/usr/bin/env python
+# coding=utf-8
+"""
+This is a script for downloading and converting the microsoft coco dataset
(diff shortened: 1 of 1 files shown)


PavlosMelissinos:
Hey there, I've been working on improving this dataset. I'm currently looking for feedback [here](#153).

Thanks

PR #90, opened 2017-06-01
save_all_weights and load_all_weights

bstriner:
Hey guys,

This is mostly just adapted from `save_model` and `load_model` in current version of Keras master `model.py`. Saves optimizer weights and model weights but not model configuration. Loads optimizer weights and model weights but you have to instantiate architecture yourself.

Functionally, `save_all_weights` goes halfway between `save` and `save_weights`.

I know this is redundant with Keras but way faster to implement here.

Please let me know your thoughts. This is my go-to method if I have something complicated that doesn't serialize well.

Cheers

### DIFF
--- a/keras_contrib/utils/save_load_utils.py
+++ b/keras_contrib/utils/save_load_utils.py
@@ -12,7 +12,7 @@ def save_all_weights(model, filepath, include_optimizer=True):
     Save model weights and optimizer weights but not configuration to a HDF5 file.
     Functionally between `save` and `save_weights`.
 
-    The saved model contains:
+    The HDF5 file contains:
         - the model's weights
         - the model's optimizer's state (if any)
     If you have a complicated model or set of models that do not serialize to JSON correctly, use this method.
@@ -80,7 +80,8 @@ def save_all_weights(model, filepath, include_optimizer=True):
 
 
 def load_all_weights(model, filepath, include_optimizer=True):
-    """Loads a model saved via `save_all_weights`.
+    """Loads the weights of a model saved via `save_all_weights`.
+    If model has been compiled, optionally load its optimizer's weights.
     # Arguments
         model: instantiated model with architecture matching the saved model.
             Compile the model beforehand if you want to load optimizer weights.
--- a/keras_contrib/utils/save_load_utils.py
+++ b/keras_contrib/utils/save_load_utils.py
@@ -1,0 +1,107 @@
+import warnings
+
+import h5py
+import keras.backend as K
+from keras import optimizers
+from keras.engine import topology
+from keras.legacy import models as legacy_models
+
+
+def save_all_weights(model, filepath, include_optimizer=True):
+    """
+    Save model weights and optimizer weights but not configuration to a HDF5 file.
+    Functionally between `save` and `save_weights`.
+
+    The saved model contains:
+        - the model's weights
+        - the model's optimizer's state (if any)
+    If you have a complicated model or set of models that do not serialize to JSON correctly, use this method.
+    # Arguments
+        model: Keras model instance to be saved.
+        filepath: String, path where to save the model.
+        include_optimizer: If True, save optimizer's state together.
+    # Raises
+        ImportError: if h5py is not available.
+    """
+    if h5py is None:
+        raise ImportError('`save_all_weights` requires h5py.')
+
+    with h5py.File(filepath, 'w') as f:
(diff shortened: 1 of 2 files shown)


keras_contrib/utils/save_load_utils.py
```diff
@@ -0,0 +1,107 @@
...
+                            dtype=val.dtype)
+                        if not val.shape:
+                            # scalar
+                            param_dset[()] = val
+                        else:
+                            param_dset[:] = val
+
+
+def load_all_weights(model, filepath, include_optimizer=True):
+    """Loads a model saved via `save_all_weights`.
```
tboquet: Maybe "loads the weights of a model saved with `save_all_weights`. 
If the model has been compiled, optionnaly load its optimizer's weights."?

keras_contrib/utils/save_load_utils.py
```diff
@@ -0,0 +1,107 @@
...
+from keras.engine import topology
+from keras.legacy import models as legacy_models
+
+
+def save_all_weights(model, filepath, include_optimizer=True):
+    """
+    Save model weights and optimizer weights but not configuration to a HDF5 file.
+    Functionally between `save` and `save_weights`.
+
+    The saved model contains:
```
tboquet: the hdf5 file contains?

bstriner: fixes are in

bstriner:
@tboquet current version has the docstring fixes, test fixes, and removes the gitkeep.

Cheers

bstriner:
@farizrahman4u please take a quick look if you get a chance. This is mostly just a copy of the save/load code from keras removing certain lines. This loads model and optimizer weights into an existing model. Keras can only load model weights into an existing model or use model and optimizer weights to instantiate a new model.

I don't like the way this code is structured. However, it mirrors keras internals exactly, so if keras has a change or a bugfix we can mirror it easily.

Cheers

keras_contrib/utils/save_load_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    if h5py is None:
+        raise ImportError('`save_all_weights` requires h5py.')
+
+    with h5py.File(filepath, 'w') as f:
+        model_weights_group = f.create_group('model_weights')
+        if legacy_models.needs_legacy_support(model):
+            model_layers = legacy_models.legacy_sequential_layers(model)
+        else:
+            model_layers = model.layers
+        topology.save_weights_to_hdf5_group(model_weights_group, model_layers)
```
xmengli999: I came across an error in this step. The error is :    RuntimeError: Unable to create attribute (Object header message is too large).

bstriner: @xmengli999 can you post some more details? Which line is actually failing? (Maybe make an issue and link it here)

xmengli999: It fails at line 35. "topology.save_weights_to_hdf5_group(model_weights_group, model_layers)". My model has a huge number of layers (around 1600). The error is : RuntimeError: Unable to create attribute (Object header message is too large). I also refer to the solution you provided in the link:
https://github.com/fchollet/keras/issues/6766
But it seems not workable in my case. Is it because my model has too many layers?

xmengli999: What is the difference of your function "save_all_weights" with the function "save_model" provided in keras?

pGit1: Any answer here? I thought this code would help me as well but it doesnt. As of now it is not possible tp save NastNet weights without significant hacking. Im still hacking and havent figured it out yet.

wollip: topology is also deprecated in keras 2.2.0 and onwards. So this needs to be replaced by saving to work. This is noted in [matterport/MASK_RCNN#566](https://github.com/matterport/Mask_RCNN/issues/566)

The fix as suggested in [matterport/MASK_RCNN#662](https://github.com/matterport/Mask_RCNN/pull/662) was to replace topology with saving.

I have also tried this function as this was the solution to [keras-team/keras$6766](https://github.com/keras-team/keras/issues/6766), however, with the fix I mentioned above as I am currently on version 2.2.2. It did not work and results in the same error from the issues and like @xmengli999 

I will look more into the difference between saving and topology to see if the fix is the cause, but it doesn't seem like it.

PR #91, opened 2017-06-01
Make callbacks into a directory

bstriner:
Hi everyone,

This PR makes `callbacks` into a directory which currently contains one file. The initializer imports everything to where it was before so no existing code should break. Just a move and rename preserving history since there is only one type of callback currently.

I was going to add a callback but I wanted to do this first.

I hope to make almost everything at the top level into a directory so it will be easier to manage adding and moving things in the future. Lets make the directory structure now so we don't end up with the gigantic files that Keras has. It would be even worse here because we have more contributors.

This just changes callbacks. Seems easiest to do it piecemeal with one PR per module so we can deal with issues on a module-by-module basis.

Any thoughts on organization?

Cheers


bstriner:
@titu1994 I had to pick a name for the submodule for your stuff and chose `keras.callbacks.snapshot`. If anyone has a clever name or something just let me know and I can update this PR.

Original PR for the snapshot callbacks: https://github.com/farizrahman4u/keras-contrib/pull/18

Cheers

titu1994:
That makes sense. Thanks for the refactor.

bstriner:
Java would force us to do one file per callback. Kind of extreme but better than the other extreme of having everything in a massive file.

Maybe 2-4 related callbacks per file is a good rule of thumb. The goal is it should be obvious from the name of the callback class what the module is and it should be obvious from the module name what the callback relates to. I'd rather err on the side of smaller files than larger files. Try reading Keras source top-to-bottom sometime.

As some examples, I have some callbacks for writing images, saving custom data, and better progress bars. I was thinking something like `keras_contrib.callbacks.image`, `save`, and `progress_bar`.

If we ever get to around 10-20 callback modules, we should talk about splitting them up into further submodules, like all of the saving related modules, etc. Don't want more than around 10 files/modules per directory.

We've got more flexibility here so we can move files around if needed. The big thing is not letting those files get too big because splitting them up is a pain especially with maintaining the history.

PR #93, opened 2017-06-01
Move DSSIMObjective to submodule

bstriner:
Making losses into directory instead of single file to make future work easier to manage. Just move two files and add one while we only have one loss to deal with.

See also https://github.com/farizrahman4u/keras-contrib/pull/91

@Dref360 it looks like you did a lot of the work on this one. Object now lives at `keras_contrib.losses.dssim.DSSIMObjective` but you can still access it using `keras_contrib.losses.DSSIMObjective` as before. Shouldn't break any existing code. Let me know if you want a different submodule name or something.

Cheers


Dref360:
LGTM

PR #94, opened 2017-06-01
Move activations, metrics, optimizers and regularizers

bstriner:
Turn four currently empty modules from single files to directories.

I moved test helpers into module itself but open to discussion on that point.

If you write a new activation, for example, add it as a submodule to `keras_contrib.activations`. Add a similarly named test file to `tests/keras_contrib/activations`. That test file at a minimum should have the standard test for that object type:

```python
from keras_contrib.tests.activations import validate_activation
from keras_contrib.activations.mymodule import MyActivation
from pytest import main

def my_activation_test():
  validate_activation(MyActivation)

if __name__=='__main__':
  main([__file__])
```

This PR puts placeholders for where those standard tests should be and standardizes some names but we can mess with those tests separately.

Cheers


the-moliver:
This is important going forward, so I'm going to merge this then work on making tests pass

PR #95, opened 2017-06-06
GloVe embeddings

LukeWoodSMU:
Merging this pull request will add a wrapper for pre-trained [GloVe Embeddings](https://nlp.stanford.edu/projects/glove/) to keras-contrib.  

The code for loading the GloVe Embedding into a keras embedding was [originally written by FChollet](https://github.com/fchollet/keras/blob/master/examples/pretrained_word_embeddings.py) I have added in cacheing and wrapped the implementation in a function.  This has made it a lot easier for me personally to test different models using a variable size GloVeEmbedding so I decided to make a pull request.

My original repository is available [here](https://github.com/lukewoodsmu/kerasglove).  

The pre trained GloVe Embeddings available come from [here](http://nlp.stanford.edu/data/glove.6B.zip).

Supported sizes are [50,100,200,300].  

**Coming in the next release:**  
- GloVeEmbeddings in from multiple data sources.


LukeWoodSMU:
I see that there are some failing tests; I will fix those and then re-open the PR.

LukeWoodSMU:
Re-opened pull request, added docstring and tests.

nzw0301:
Are there some specific function of `GloveEmbedding`?

LukeWoodSMU:
GloVe is an unsupervised algorithm for creating dense vector representations of words.  You can use these to prevent the need for taking the huge amount of time required to train your own embedding.  

Training your own embedding makes the number of trainable parameters in a model ridiculously large causing long training times.

You get much stronger representations using pre-trained GloveEmbeddings.

If you are asking about my specific implementation then the answer is no, but it is much easier to construct a GloveEmbedding using my wrapper than it is from scratch.  Assuming you already have the embedding downloaded here is the code required to get a pre-trained embedding:

`
    EMBED_SIZE = int(size)

    fname = "glove.6B.300d.txt"

    embeddings_index = {}
    for line in open(fname):
        values = line.split()
        word = values[0]
        coefs = np.asarray(values[1:],dtype='float32')
        embeddings_index[word] = coefs

    embedding_matrix=np.zeros((len(word_index)+1,EMBED_SIZE))
    for word,i in word_index.items():
        vec = embeddings_index.get(word)
        if vec is not None:
            embedding_matrix[i] = vec

    embedding_layer = Embedding(
            len(word_index)+1,
            EMBED_SIZE,
            weights=[embedding_matrix],
            input_length=input_length,
            trainable=False,**kwargs)`

I decided this was really repetitive for people who use lots of RNNs and decided to create this wrapper as a result of that.

TLDR:  The benefit of using pre-trained embeddings is that you can get much more valuable dense vector representations of text data while training less variables.  You can already get this functionality out of keras, it just requires ~20ish lines of code every time you want to do it and requires you to have already downloaded the embedding.

nzw0301:
You also said that this class doesn’t contain some new attribute and `call` function in `Embedding` layer, just loading function to Glove's word embeddings and pass it to `weights` argument of `Embedding`.

I recommend creating a new `utils` function to load any pre-training embedding weights include Glove and word2vec/fastText format, instead of creating the new class for loading pre-trained Glove embeddings for more Keras users.

Issue #63, opened 2017-03-31
DenseNetFCN not training to expected performance


ahundt:
## I'm training and testing DenseNetFCN on Pascal VOC2012

Could I get advice on next steps to take to debug and improve the results?

To do so I'm using the [train.py](https://github.com/ahundt/Keras-FCN/blob/densenet_atrous/train.py) training script in [my fork of Keras-FCN](https://github.com/ahundt/Keras-FCN/blob/densenet_atrous/) in conjunction with [DenseNetFCN in keras-contrib](https://github.com/ahundt/keras-contrib/blob/64630aed0386f33db592ad3ccc9265967fc87d65/keras_contrib/applications/densenet.py#L241) with #46 applied, which for DenseNetFCN really mostly changes the formatting for pep8, though the regular densenet is more heavily modified. 

 I use Keras-FCN because we don't have an FCN training script here in keras-contrib yet, though I plan to adapt & submit this here once things work properly. While the papers don't publish results on Pascal VOC, the [original DenseNet](https://arxiv.org/pdf/1608.06993.pdf) has fairly state of the art results on ImageNet & cifar10/cifar100, and [DenseNetFCN](https://arxiv.org/pdf/1611.09326.pdf) performed well on CamVid and the Gatech dataset. Considering this I expected that perhaps DenseNetFCN might fail to get state of the art results, but I figured it could most likely in a worst case get over 50% mIOU and around 70-80% pixel accuracy since it has many similarities in common with ResNet, and performed quite well on the much smaller CamVid dataset.

### DenseNet FCN configuration I'm using

```python
    return densenet.DenseNetFCN(input_shape=(320, 320, 3),
                                weights=None, classes=classes,
                                nb_layers_per_block=4,
                                growth_rate=13,
                                dropout_rate=0.2)
```

This is very close to the configuration FC-DenseNet56 from the [100 layers tiramisu aka DenseNetFCN paper](https://arxiv.org/pdf/1611.09326.pdf).

#### Sparse training accuracy

Here is what I'm seeing as I train with the `Adam` optimizer and loss rate of `0.1`:

```
lr: 0.100000
Epoch 2/450
366/366 [==============================] - 287s - loss: 62.5131 - 
sparse_accuracy_ignoring_last_label: 0.3657
[...snip...]
lr: 0.100000
Epoch 148/450
366/366 [==============================] - 286s - loss: 82.2138 - 
sparse_accuracy_ignoring_last_label: 0.3375
```

[...]

PR #96, opened 2017-06-06
Fix training difficulty for DenseNetFCN

titu1994:
It was brought to my attention, that there was a connection missing in the final layer of the DenseNetFCN, which was the reason for the poor performance on other datasets. (https://github.com/titu1994/DenseNet/pull/15)

This has been tested on Keras 2.0.4+, and I can confirm that the model is performing far better than before on a private medical dataset. 

Fixes https://github.com/farizrahman4u/keras-contrib/issues/63
@ahundt, this may be of use to you.

### DIFF
--- a/keras_contrib/applications/densenet.py
+++ b/keras_contrib/applications/densenet.py
@@ -178,7 +178,7 @@ def DenseNet(input_shape=None, depth=40, nb_dense_block=3, growth_rate=12, nb_fi
 def DenseNetFCN(input_shape, nb_dense_block=5, growth_rate=16, nb_layers_per_block=4,
                 reduction=0.0, dropout_rate=0.0, weight_decay=1E-4, init_conv_filters=48,
                 include_top=True, weights=None, input_tensor=None, classes=1, activation='softmax',
-                upsampling_conv=128, upsampling_type='upsampling'):
+                upsampling_conv=128, upsampling_type='deconv'):
     '''Instantiate the DenseNet FCN architecture.
         Note that when using TensorFlow,
         for best performance you should set
@@ -216,7 +216,7 @@ def DenseNetFCN(input_shape, nb_dense_block=5, growth_rate=16, nb_layers_per_blo
             activation: Type of activation at the top layer. Can be one of 'softmax' or 'sigmoid'.
                 Note that if sigmoid is used, classes must be 1.
             upsampling_conv: number of convolutional layers in upsampling via subpixel convolution
-            upsampling_type: Can be one of 'upsampling', 'deconv' and
+            upsampling_type: Can be one of 'deconv', 'upsampling' and
                 'subpixel'. Defines type of upsampling algorithm used.
             batchsize: Fixed batch size. This is a temporary requirement for
                 computation of output shape in the case of Deconvolution2D layers.
@@ -382,10 +382,10 @@ def __dense_block(x, nb_layers, nb_filter, growth_rate, bottleneck=False, dropou
     x_list = [x]
 
     for i in range(nb_layers):
-        x = __conv_block(x, growth_rate, bottleneck, dropout_rate, weight_decay)
-        x_list.append(x)
+        conv_block = __conv_block(x, growth_rate, bottleneck, dropout_rate, weight_decay)
+        x_list.append(conv_block)
 
-        x = concatenate(x_list, axis=concat_axis)
+        x = concatenate([x, conv_block], axis=concat_axis)
 
         if grow_nb_filters:
             nb_filter += growth_rate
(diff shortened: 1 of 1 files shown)


titu1994:
@farizrahman4u Could you review this PR and merge?

ahundt:
wow great! let me test before we say #63 is fully fixed

titu1994:
Yup that's true. But when I tested it on a private dataset it trained properly and didn't get stuck as before. 

Even considering that the number of parameters exactly matches the ones in the paper is a sign of hope.

titu1994:
@farizrahman4u @the-moliver Could you review and merge please?

ahundt:
I'm still training, it is just about to pass the accuracy threshold that can be reached with labeling every pixel background:
```
Epoch 56/250
 6098/11127 [===============>..............] - ETA: 2170s - loss: 2.3599 - sparse_accuracy_ignoring_last_label: 0.7199 
```

I think this is a good sign!

titu1994:
Mine finished training, finally getting better results than the UNet baseline that I was using. 

On a side note, are you training that model on a CPU? Or does it take that long even on a GPU/s? If it takes that long on a GPU, it must be one of the largest models I've ever seen.

ahundt:
It is on a gpu, I'm not sure why it has been so slow... While it is the largest of the models from the tiramisu paper, it is not actually that large. It might be obvious if I profile it lol.

keras_contrib/applications/densenet.py
```diff
@@ -178,7 +178,7 @@ def DenseNet(input_shape=None, depth=40, nb_dense_block=3, growth_rate=12, nb_fi
 def DenseNetFCN(input_shape, nb_dense_block=5, growth_rate=16, nb_layers_per_block=4,
                 reduction=0.0, dropout_rate=0.0, weight_decay=1E-4, init_conv_filters=48,
                 include_top=True, weights=None, input_tensor=None, classes=1, activation='softmax',
-                upsampling_conv=128, upsampling_type='upsampling'):
+                upsampling_conv=128, upsampling_type='deconv'):
```
titu1994: Because it matches the paper. Also, I'm getting slight better results with it than with upsampling or subpixel.

keras_contrib/applications/densenet.py
```diff
@@ -382,10 +382,10 @@ def __dense_block(x, nb_layers, nb_filter, growth_rate, bottleneck=False, dropou
     x_list = [x]
 
     for i in range(nb_layers):
-        x = __conv_block(x, growth_rate, bottleneck, dropout_rate, weight_decay)
-        x_list.append(x)
+        conv_block = __conv_block(x, growth_rate, bottleneck, dropout_rate, weight_decay)
+        x_list.append(conv_block)
 
-        x = concatenate(x_list, axis=concat_axis)
+        x = concatenate([x, conv_block], axis=concat_axis)
```
titu1994: This is the important fix that I mentioned. With it, and Deconvolution upsampling, the parameter count exactly matches the paper and trains properly. 

The previous version had fewer parameters, and did not train properly.

ahundt: My version is ever so slightly different from both of your versions. I think my version may be slightly more accurate w.r.t. the paper's description. That said, it is also possible that this change of yours could also outperform the original paper. :-)

I looked carefully at the following before writing the code above:
- your previous version
- your current version
- the paper and its [corresponding code](https://github.com/SimJeg/FC-DenseNet/blob/master/FC-DenseNet.py#L67)

Note that in the version I posted `x_list` and `return_concat_list` are gone. I may very well still be mistaken here, but could you either correct me or confirm my suspicion?

titu1994: Sorry for the late reply. 

Without the return of concat list, the upsample path will not be correct anymore. As in, the upsample block concatenates a long term dependency, and that is provided by the concat_list.

I tried modifying the code to work without that, but there are fewer number of parameters than the original paper and the current implementation, and that suggests that it may suffer during training.

ahundt: Okay that's evidence enough for me! As a side test, I could try running those changes if you still have them somewhere.

titu1994: Ah I reverted my code of the changes after I was done. Sorry about that.

ahundt: oh well, think it is worthwhile?

ahundt:
@farizrahman4u can this be merged?

titu1994:
@farizrahman4u @the-moliver Could you merge this please? 

I've finished with MobileNets and ResNeXt, but I can't add PRs until this one gets merged (I made changes to master in my branch, and new branches would include this PR as well)

PR #97, opened 2017-06-07
Instance Normalization layer

mathieumb:
Instance Normalization layer (Lei Ba et al, 2016, Ulyanov et al., 2016), which is in the list of requested contributions.
Individually normalize each instance in the batch.

I called it `InstanceNormalization` because:
-Its a named used for it in the literature
-Simply `Normalization` would be a headache when searching for it
-`LayerNormalization layer` doesn't sound right.


mathieumb:
Will reopen after fixing theano-specific issue

PR #101, opened 2017-06-08
Instance Normalization layer

mathieumb:
Instance Normalization layer (Lei Ba et al, 2016, Ulyanov et al., 2016), which is in the list of requested contributions.
Individually normalize each instance in the batch.

I called it `InstanceNormalization` because:
-Its a named used for it in the literature
-Simply `Normalization` would be a headache when searching for it
-`LayerNormalization` layer doesn't sound right.


keras_contrib/layers/normalization.py
```diff
@@ -6,6 +6,141 @@
...
+        (tuple of integers, does not include the samples axis)
+        when using this layer as the first layer in a model.
+    # Output shape
+        Same shape as input.
+    # References
+        - [Layer Normalization](https://arxiv.org/abs/1607.06450)
+        - [Instance Normalization: The Missing Ingredient for Fast Stylization](https://arxiv.org/abs/1607.08022)
+    """
+    def __init__(self,
+                 axis=-1,
```
bstriner: Should the default be None?

keras_contrib/layers/normalization.py
```diff
@@ -6,6 +6,141 @@
...
+        else:
+            self.beta = None
+        self.built = True
+
+    def call(self, inputs, training=None):
+        input_shape = K.int_shape(inputs)
+        reduction_axes = list(range(1, len(input_shape)))
+
+        if (self.axis is not None):
+            del reduction_axes[self.axis]
```
bstriner: The range starts at 1, so axis=1 will remove the third dim from the reduction.

keras_contrib/layers/normalization.py
```diff
@@ -6,6 +6,141 @@
...
+
+        mean = K.mean(inputs, reduction_axes, keepdims=True)
+        stddev = K.std(inputs, reduction_axes, keepdims=True) + self.epsilon
+        normed = inputs / stddev + (- mean / stddev)
```
bstriner: Why not `(inputs-mean)/stddev`? Seems simpler unless there is some other reason.

keras_contrib/layers/normalization.py
bstriner: (Docs should be clear about what axis is relative to)

bstriner:
Does it make sense use a running std and mean like batchnorm or do you think test data should normalize test data?
Does it make sense to break out the scale/centering code into a utility or a separate layer? I could see those being useful by themselves. Not sure if the axes for scaling and centering should be tied to the axes for normalization. Could be two separate layers. Could also be some utility functions used by this layer but also used in other layers.

Cheers

mathieumb:
@bstriner 

Thanks for the review, im fixing the code and adding a new unit test to address your comments.

I believe we should not have a running std and mean, one of the 'features' of Instance Normalization being that it does the same thing at training and testing time (unlike Batch Normalization)

I think the scaling/centering is a bit too simple to break into an utility, and in my opinion its better to leave it in the layer as in most cases you will want it. As a separate layer Im not convinced they have a use unless preceded by one of the normalization layer, but maybe thats just ignorance on my part?

keras_contrib/layers/normalization.py
mathieumb: I left it at -1 to keep it uniform with the other Normalization layers

bstriner: Default should be whatever works with standard 2d dense data. Am I wrong that `None` is what you would use in that situation? If there are 2 dims, it looks like -1 would mean `reduction_dims = []` but `None` means `reduction_dims=[1]` which is what you want. If we want -1 to be the default, gotta change what -1 means to work for a standard dense layer.

mathieumb:
Ok, so I changed the axis default to None. Axis is in the whole input frame (Axis=0 is the batch (illegal), Axis=1 is the first input dimension, etc). I updated the doc to made it clear and added a new unit test that shows & verify its use. Unit tests should pass when master does.

keras_contrib/layers/normalization.py
```diff
@@ -6,6 +6,145 @@
...
+                                        initializer=self.beta_initializer,
+                                        regularizer=self.beta_regularizer,
+                                        constraint=self.beta_constraint)
+        else:
+            self.beta = None
+        self.built = True
+
+    def call(self, inputs, training=None):
+        input_shape = K.int_shape(inputs)
+        reduction_axes = list(range(0, len(input_shape)))
```
the-moliver: can't this be:
```reduction_axes = list(range(1, len(input_shape)))```

File: keras_contrib/layers/normalization.py [hunk 5]
```diff
@@ -6,6 +6,145 @@
...
+
+        if (self.axis is not None):
+            del reduction_axes[self.axis]
+
+        del reduction_axes[0]
```
the-moliver: and this line then removed

keras_contrib/layers/normalization.py
mathieumb: Hi, as @bstriner pointed out before I changed it, in this case the line 109 would not remove the correct axis (if positive)

the-moliver:
Ah good point!

PR #104, opened 2017-06-10
add Scaled Exponential Linear Unit activation

rilut:
Thanks


titu1994:
The SELU paper has another variant called selu_dropout which is also important. It should be added as well to this PR.

farizrahman4u:
Closing redundant PR.

PR #110, opened 2017-06-16
FTML optimizer

szhengac:
Hi, this is an implementation of deep learning optimizer [FTML](http://www.cse.ust.hk/~szhengac/papers/icml17.pdf). 

This PR is directed from [fchollet/keras#7004](https://github.com/fchollet/keras/pull/7004). Thanks.


the-moliver:
@szhengac I just pushed tests for your optimizer here https://github.com/farizrahman4u/keras-contrib/commit/ef6c77063954bcb3c363d7d0b264cbc3d3dc801b. Since it wasn't fully set up I went ahead and did myself, but next time please include tests with a PR. Also fixed a small bug where you forgot to include the `self.epsilon = epsilon` in the `__init__` function. Thanks again for the PR, look forward to playing with it!

szhengac:
I just note that all the optimizers in keras-1 do not have `self.epsilon = epsilon`. I will add a comparison with Adam in the future.  In the following, I attach the training results I obtained before using the `cifar10_cnn.py` without data augumentation in Keras examples. Here, the validation per epoch was performed on the training set, since I wanted to check the training loss over the whole training set. The testing was done after the training.

FTML:
ftml
ftml_test

Adam:
adam
adam_test

Issue #102, opened 2017-06-09
Adding ResNeXt as an application


titu1994:
Hi, I was wondering if it is alright to add ResNeXt model building code from the paper [Aggregated Residual Transformations for Deep Neural Networks](https://arxiv.org/pdf/1611.05431.pdf) without pre-trained weights for CIFAR-10 or CIFAR100.

I'm currently using my GPU for another research project, but since ResNeXt models aren't currently available for Keras, I was hoping to add them to contrib so that someone else could use it to train on CIFAR or even ImageNet.

We can update the code later on to point to the weights for CIFAR.

the-moliver:
Sure seems reasonable to me since there is probably interest.

hasibzunair:
@titu1994 were the resnext implementation on cifar 10 and 100 added in here?

titu1994:
Sadly, they weren't. Keras, and Tensorflow do not support Grouped Convolutions yet, and my hacky workaround reduces performance far too much to be of any value.

hgaiser:
I guess this situation has remained unchanged in the past 22 days? I find it odd that grouped convolutions is not added to tensorflow yet. Clearly it has a purpose, considering the state-of-the-art uses it and CUDA / cuDNN support it natively now.

I'd be curious to hear if you find a way to implement ResNeXt in an efficient manner, please let me know if you have any updates.

PGryllos:
@titu1994 do you think there is interest for a backend implementation of grouped convolutions? for us resnext looks like it works very well, but training times are prohibitive :/

gabrieldemarmiesse:
New layers are welcome in keras contrib. You can also do pull requests for new backend operation.

titu1994:
Grouped Convolutions are more efficient than the hacky solution I put together, since they are implemented at cudnn or Cuda level in PyTorch and original Torch Lua. 

A Keras backend function still needs to utilize ops provided by the underlying engine, here - Tensorflow, Theano or CNTK, to perform an operation. 

Since neither of the three backends support Grouped Convolutions directly, there isn't going to be much of a speed benefit. We can only hope TF 2 or CNTK will eventually support Grouped Convolutions in the future.

PGryllos:
[...]

PR #111, opened 2017-06-17
Adds ResNeXt model builders

titu1994:
Adds the ResNeXt models from the paper [Aggregated Residual Transformations for Deep Neural Networks](https://arxiv.org/pdf/1611.05431.pdf)

Adds 2 types of ResNext models-
- General style (ResNext)
- Imagenet optimized (ResNextImageNet)

Weight paths need to be added, after than, it will support ResNext-29(8x64d) for general style model and ResNext-50(32x4d) for ImageNet optimized model.

### DIFF
--- a/keras_contrib/applications/__init__.py
+++ b/keras_contrib/applications/__init__.py
@@ -1,2 +1,4 @@
 from .densenet import DenseNet
 from .ror import ResidualOfResidual
+from .wide_resnet import WideResidualNetwork
+from .resnext import ResNext, ResNextImageNet
--- a/keras_contrib/applications/resnext.py
+++ b/keras_contrib/applications/resnext.py
@@ -1,6 +1,6 @@
 '''ResNeXt models for Keras.
 # Reference
-- [Densely Connected Convolutional Networks](https://arxiv.org/pdf/1608.06993.pdf))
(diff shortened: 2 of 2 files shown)


titu1994:
@the-moliver @farizrahman4u PR is ready for review. As discussed in https://github.com/farizrahman4u/keras-contrib/issues/102, I have added support for weights, but since I don't have nearly enough computational power to train these models, I am leaving the weight paths as empty strings.

I see that facebook has posted weights in https://github.com/facebookresearch/ResNeXt. However, since I am on windows, Torch and Caffe don't work on it, so I can't port the weights myself.

Even if I could, Keras does not support Grouped Convolutions. That is the reason I implemented the equivalent block [2] from the paper. This leads to exact same number of parameters, but more complex design. 

Grouped convolutions support would provide nice syntactic sugar, to say the least.

the-moliver:
It looks like grouped convolutions are coming soon to Theano at least, so maybe it would be good to move your group convolution implementation to the backends, so that it could be easily replaced. Check https://github.com/Theano/Theano/pull/5991 for details.

titu1994:
Hmm if that is the case, I think I will hold back on the PR for now. Once both tf and th (hopefully CNTK as well) support it, I will update my repo and then submit a PR.

PR #115, opened 2017-06-28
Added Dead ReLU detector callback

arodiss:
Added new callback which reports the number of dead ReLUs after each training epoch
ReLU is considered to be dead if it did not fire once for entire training set

### DIFF
--- a/keras_contrib/callbacks/__init__.py
+++ b/keras_contrib/callbacks/__init__.py
@@ -1,1 +1,2 @@
 from .snapshot import SnapshotCallbackBuilder, SnapshotModelCheckpoint
+from .dead_relu_detector import DeadReluDetector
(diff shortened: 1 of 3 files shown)


arodiss:
`save_load_utils_test.py::test_save_and_load_all_weights()` is failing even on master: https://github.com/farizrahman4u/keras-contrib/pull/116

infelane:
Hi, I updated the dead_relu_detector.py on the following things:
1) Extended layers to everything that has a relu activation
2) Added layer name to the warning
3) Corrected the dead neuron count (should work for both Theano and TF (tested for TF))
4) Added a print alongside the warning since for me the warning stopped showing after some epochs (might be a bug at my side).

However this is my first time working on public github repositories. I'll update if I succeeded in pushing.
I also don't know how I have to work with tests.

arodiss:
@infelane 
Can you please provide link to your pull request?

infelane:
@arodiss 
Link to pull request: https://github.com/farizrahman4u/keras-contrib/pull/174

arodiss:
Thank you @infelane

PR #118, opened 2017-07-05
Separable Fully Connected Layer

alexandari:
Separable Fully Connected Layers (Paper https://doi.org/10.1101/146431 accepted to ICML2017 Workshop on Comp Bio)

It is a module for deep neural networks specific to genomic sequences. This is a decomposable fully connected layer (separable fully-connected layers, abbrv. SeparableFC) as well as a second difference (curvature) regularizer and constraint. The layer separates positional patterns from channel dependent patterns, allowing each to be learned independently.  This in turn leads to a more efficient and interpretable modeling of complex patterns in genomic data. The regularizer and constraint can be readily applied to the tensor concerned with the positional weights.


alexandari:
Hi @farizrahman4u @the-moliver 

Just touching base for a quick update on this because at least a few people asked us for useable Keras extensions instead of our lab's fork. Please let us know.

Many thanks!

PR #120, opened 2017-07-17
Adding PPV and NPV to the metrics list

kmader:
adding positive and negative predictive value to metrics (can also be used as losses or combined with other functions in losses)


kmader:
combined with https://github.com/farizrahman4u/keras-contrib/pull/121

PR #122, opened 2017-07-18
Implementation of mc_dropput

javiercorrea:
- Implemented mc_dropout parameters for the Dropout layer
  allowing to Dropout layer to drop inputs at prediction
  time

### DIFF
--- a/keras_contrib/layers/core.py
+++ b/keras_contrib/layers/core.py
@@ -169,3 +169,57 @@ class CosineDense(Layer):
 
 
 get_custom_objects().update({'CosineDense': CosineDense})
+
+class Dropout(Layer):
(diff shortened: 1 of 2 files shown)


alexandari:
I don't think the specialized layer is necessary. A simpler way to get uncertainty estimates is to set learning_phase = 1 as outlined in the FAQ: https://keras.io/getting-started/faq/#how-can-i-obtain-the-output-of-an-intermediate-layer

gabrieldemarmiesse:
Even easier, when calling the normal dropout, you can do this:

`x = Dropout(0.5)(x, training=True)` and tada! The dropout layer will always behave as if it was in training mode.

PR #124, opened 2017-07-19
Adding hyperparameter search utils to Keras.

dnola:
Hello,

In this PR I add a basic framework we can build on to automatically tune hyperparameters in Keras, while allowing models not being actively trained to stay on disk without disrupting callbacks. It is fairly straightforward, check the example code to see how it might work.

There still isn't a straightforward way to tune hyperparameters for graphical models in Keras (or any model period, really) - this PR is to address that. Furthermore, this is critical (and simple) enough that we shouldn't need to require an entire scikit-learn dependency, wrapper, and workflow to do hyperparameter tuning.

Additionally, we probably need better examples of how to tune hyperparameters automatically in Keras. The example code in this PR provides this. 

Lastly, relying on scikit-learn - which isn't Keras aware - to perform our hyperparameter search can cause lots of unexpected problems. As such, this approach never quite took off. Failure to reset the Tensorflow computational graph, keeping too many models in memory at once, performing unsafe copies of models, callbacks don't work, Model() doesn't work, etc.

This is just a useful starting point. I intend on continuing to build on this to provide even simpler Random and Grid search interfaces, as well as simple interfaces to deep-learning-focused search techniques like Hyperband and maybe Harmonica. The current interface is granular enough to support most hyperparameter optimization workflows for now, I think.

Tested in Python 2.7 and 3.6 on Theano and Tensorflow. 

Let me know what you think, Thanks!


ahundt:
How does this compare to hyperas? https://github.com/maxpumperla/hyperas

dnola:
Thanks for your comment!

For now it is a starting point for some higher level stuff in the vein Hyperas. A few key issues with most tuning frameworks (including the sklearn wrapper, and Hyperas) is that they don't play well with adaptive resource allocation. That is, when models need to be paused and resumed, or allocated fractions of an epoch, they fall apart - the callbacks especially, but memory consumption is also an issue. This is important because state of the art for most DL HPO right now seems to belong to adaptive allocation strategies like Hyperband or Harmonica, where you need to hold on to a bunch of partially trained candidates and allocate more resources at the end of a round. This trend towards adaptive allocation is likely to continue as the Bayesian and Bandit approaches get mixed, as they aren't mutually exclusive. 

Part of this is really to serve as an alternative to the sklearn wrapper that is all Keras - and doesn't take dependencies outside what Keras already does. 

It's just a starting point to enable some of the more modern HPO approaches in a high level way. Happy to take feedback or build on it.

dnola:
The example code does a good job of demonstrating this, as a note. It is a nice simple example of adaptive allocation. I tried to make it pretty easy to read - let me know what you think! It is basically "survival of the fittest". These are low level primitives though - I'm hoping to make this a little higher level for people who want more of a "one liner" approach.

ahundt:
I finally got a chance to look through this and overall it is neat! Am I correct that it only implements random search at the moment?

Issue #127, opened 2017-07-29
BatchRenormalization uses Numpy code where it should use TF code


waleedka:
Looking at the code of [BatchRenormalization ](https://github.com/farizrahman4u/keras-contrib/blob/master/keras_contrib/layers/normalization.py), I think there is a serious issue there. This section, for example, inside the call() function:
```
# update r_max and d_max
t_val = K.get_value(self.t)
r_val = self.r_max_value / (1 + (self.r_max_value - 1) * np.exp(-t_val))
d_val = self.d_max_value / (1 + ((self.d_max_value / 1e-3) - 1) * np.exp(-(2 * t_val)))
t_val += float(self.t_delta)
```

The main issue is that the code uses Python/Numpy operations, rather than Keras/TensorFlow operations. Note that the layer call() function is executed only once when the model is created. It does *not* get called repeatedly for every batch. The call() function should create a TF/Theano computation graph rather than do the computation directly in Python/Numpy.

What's happening there is that it calls K.get_value(), which internally creates a TF session and runs it to get the float value of ```self.t```. Then, r_val and d_val and calculated in Python code and used in subsequent update ops. Since r_val and d_val are computed only once when the model is created, they'll have values of 1 and 0 respectively and will remain so. In other words, if I'm reading this correctly, this layer is operating as a regular batch normalization layer, not as a batch re-normalization layer. 

I didn't run the code to confirm so I could be missing something. If so, I'd appreciate being corrected.

stygstra:
[Another place](https://github.com/farizrahman4u/keras-contrib/blob/c4d1a617461b0a4e29ec087de562032d47d03244/keras_contrib/layers/normalization.py#L293-L299) where `K.get_value` is used:
```python
        r_max_value = K.get_value(self.r_max)
        r = std_batch / (K.sqrt(self.running_variance + self.epsilon))
        r = K.stop_gradient(K.clip(r, 1 / r_max_value, r_max_value))

        d_max_value = K.get_value(self.d_max)
        d = (mean_batch - self.running_mean) / K.sqrt(self.running_variance + self.epsilon)
        d = K.stop_gradient(K.clip(d, -d_max_value, d_max_value))
```

Since this is in call(), the clipping is fixed at `r_max_value=1` and `d_max_value=0` rather than smoothly approaching `r_max_value=3` and `d_max_value=5` as described in section 4, paragraph 3 of the [batch renorm paper](https://arxiv.org/pdf/1702.03275.pdf).

[...]

PR #128, opened 2017-07-29
change np.exp to K.exp

the-moliver:
To address https://github.com/farizrahman4u/keras-contrib/issues/127

### DIFF
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -318,14 +318,12 @@ class BatchRenormalization(Layer):
                          K.moving_average_update(self.running_variance, std_batch ** 2, self.momentum)], inputs)
 
         # update r_max and d_max
-        t_val = K.get_value(self.t)
-        r_val = self.r_max_value / (1 + (self.r_max_value - 1) * K.exp(-t_val))
-        d_val = self.d_max_value / (1 + ((self.d_max_value / 1e-3) - 1) * K.exp(-(2 * t_val)))
-        t_val += float(self.t_delta)
+        r_val = self.r_max_value / (1 + (self.r_max_value - 1) * K.exp(-self.t))
+        d_val = self.d_max_value / (1 + ((self.d_max_value / 1e-3) - 1) * K.exp(-(2 * self.t)))
 
         self.add_update([K.update(self.r_max, r_val),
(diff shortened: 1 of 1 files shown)


keras_contrib/layers/normalization.py
```diff
@@ -319,8 +319,8 @@ def call(self, inputs, training=None):
 
         # update r_max and d_max
         t_val = K.get_value(self.t)
```
waleedka: I believe the real problem is here. K.get_value() runs a TensorFlow session internally to read the value of self.t and returns a float value. This value is 0 when you create the model. And since the call() function is called only once, then t_val will always be zero, and the following calculations will always be based on a t_val value of 0

titu1994:
@waleedka Hi, I am the author of the BatchRenorm code. This was always a hacky solution to simulate time per iteration. 

While the call function was called only once, I added an update to the self.t value (which is updated per batch, thus simulates time step). However, there was another bug where I would constantly update the value in numpy. Since this wasnt in tensorflow graph, the value would stay set to `self.t += float(self.t_delta)`, which turns out to be 1 (since t_delta = 1) and would constantly be updated to same value.

With self.t = 1, this simulates batch renorm behaviour (though not exactly like the paper suggests). This was as close to a fix as I could find : 
```
# update r_max and d_max
r_val = self.r_max_value / (1 + (self.r_max_value - 1) * K.exp(-self.t))
d_val = self.d_max_value / (1 + ((self.d_max_value / 1e-3) - 1) * K.exp(-(2 * self.t)))

self.add_update([K.update(self.r_max, r_val),
                            K.update(self.d_max, d_val),
                            K.update_add(self.t, K.variable(np.array([self.t_delta])))], x)
```

Now, the entire graph is in TF, and the update to self.t is an update_add of a constant value (self.t_delta).

One major change needed for this to work is to reduce t_delta value to 1e-3 or 2e-3. That causes batch renorm to act properly as before.

The fact that the renorm version of WRN-16-4 converges to 61% in the 1st epoch, whereas the batchnorm version only gets 60~% after the first 5 epochs suggests that renorm was working before and will work now as well.

@the-moliver Could you merge these changes in this PR itself? I am travelling and wont be able to create a PR for several days.

PR #137, opened 2017-08-17
Update ftml for Keras changes to optimizers

the-moliver:


joelkuiper:
I now get "AttributeError: module 'keras.legacy.interfaces' has no attribute 'legacy_get_updates_support'" with Keras 2.0.6, is this expected? Thanks!

ahundt:
@joelkuiper keras-contrib only tries to support the latest release and master. If you're able to try updating your keras version.

PR #142, opened 2017-08-30
Fix BatchRenormalization clipping (fixes #127)

### LINKED ISSUE #127: BatchRenormalization uses Numpy code where it should use TF code

stygstra:
Fixed the `r_max` and `d_max` clipping schedule. It was previously fixed at r_max=1 and d_max=0, making it essentially equivalent to regular batch normalization. See [my comment](https://github.com/farizrahman4u/keras-contrib/issues/127#issuecomment-325903950) in #127.

(Edit 2017-10-10: Removed parts that @titu1994 fixed in #151.)

### DIFF
--- a/keras_contrib/backend/cntk_backend.py
+++ b/keras_contrib/backend/cntk_backend.py
@@ -1,2 +1,26 @@
 from keras.backend import cntk_backend as KCN
 import cntk as C
+import numpy as np
+
+
+def clip(x, min_value, max_value):
+    """Element-wise value clipping.
+
+    If min_value > max_value, clipping range is [min_value,min_value].
+
+    # Arguments
+        x: Tensor or variable.
+        min_value: Tensor, float, int, or None.
+            If min_value is None, defaults to -infinity.
+        max_value: Tensor, float, int, or None.
+            If max_value is None, defaults to infinity.
+
+    # Returns
+        A tensor.
+    """
+    if max_value is None:
+        max_value = np.inf
+    if min_value is None:
+        min_value = -np.inf
+    max_value = C.maximum(min_value, max_value)
+    return C.clip(x, min_value, max_value)
--- a/keras_contrib/backend/tensorflow_backend.py
+++ b/keras_contrib/backend/tensorflow_backend.py
@@ -1,4 +1,5 @@
 import tensorflow as tf
+import numpy as np
 
 try:
     from tensorflow.python.ops import ctc_ops as ctc
@@ -11,6 +12,7 @@ from keras.backend.tensorflow_backend import _postprocess_conv3d_output
 from keras.backend.tensorflow_backend import _preprocess_padding
 from keras.backend.tensorflow_backend import _preprocess_conv2d_input
 from keras.backend.tensorflow_backend import _postprocess_conv2d_output
+from keras.backend.tensorflow_backend import _to_tensor
 
 py_all = all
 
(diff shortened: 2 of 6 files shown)


keras_contrib/backend/tensorflow_backend.py
```diff
@@ -158,3 +160,22 @@ def moments(x, axes, shift=None, keep_dims=False):
     ''' Wrapper over tensorflow backend call '''
 
     return tf.nn.moments(x, axes, shift=shift, keep_dims=keep_dims)
+
+
+def clip(x, min_value, max_value):
```
stygstra: This was modified because the `K.clip` in Keras doesn't allow tensors for `min_value` or `max_value`.

titu1994:
Hmm I corrected a few of these fixes in my PR. However, I didn't need any of these clip additions.

stygstra:
@titu1994 I'll rebase onto master. The clipping-related changes address the issue noted in [this comment](https://github.com/farizrahman4u/keras-contrib/issues/127#issuecomment-325903950).

keras_contrib/backend/cntk_backend.py
```diff
@@ -1,2 +1,10 @@
 from keras.backend import cntk_backend as KCN
 import cntk as C
+import numpy as np
+
+
+def clip(x, min_value, max_value):
+    if max_value is None:
```
ahundt: if max_value defaults to inf, should min default to negative infinity?

stygstra: You're correct. The Keras CNTK backend uses -inf as the default min_value. I did it this way because the TensorFlow backend *doesn't* default to -inf for min_value ([ref](https://github.com/fchollet/keras/blob/2.0.8/keras/backend/tensorflow_backend.py#L1472-L1489)), but failed to notice that the behavior of the CNTK backend differed.

ahundt:
@stygstra Could you resolve the conflicts here?

keras_contrib/backend/cntk_backend.py
stygstra: I changed my mind and extended all backends to allow `None` for `min_value`. See notes in `tests/keras_contrib/backend/backend_test.py` for rationale and details.

stygstra:
@ahundt Conflicts resolved. I changed the behavior of K.clip slightly since the previous iteration. See my comment in the review.

keras_contrib/backend/tensorflow_backend.py
ahundt: Documentation needs to be updated accordingly. Could these changes be valuable in keras itself?

keras_contrib/backend/cntk_backend.py
ahundt: Documentation needs to be updated accordingly. Could these changes be valuable in keras itself?

keras_contrib/backend/tensorflow_backend.py
stygstra: @ahundt Updated docstrings for all 3 backends. I think these changes could be useful in Keras for people writing new layers.

PR #143, opened 2017-09-01
Correct dense connectivity in DenseNet

titu1994:
Due to an earlier commit, there was an incorrect update to the dense_block. 
It had no dense connectivity at all, even though it had the same number of parameters.

This patch fixes that.

Edit : Interestingly, the models are exactly equivalent (not in operation, but in performance). As in if you load either format of dense connectivity, the weights will load (if you use by_name=True) and the model will give exact same classification results. However, in the correct case, there is dense connectivity and the incorrect case it just looks like a resnet.

### DIFF
--- a/keras_contrib/applications/densenet.py
+++ b/keras_contrib/applications/densenet.py
@@ -385,7 +385,7 @@ def __dense_block(x, nb_layers, nb_filter, growth_rate, bottleneck=False, dropou
         conv_block = __conv_block(x, growth_rate, bottleneck, dropout_rate, weight_decay)
         x_list.append(conv_block)
 
-        x = concatenate([x, conv_block], axis=concat_axis)
+        x = concatenate(x_list, axis=concat_axis)
 
         if grow_nb_filters:
             nb_filter += growth_rate


stygstra:
densenet before/after

titu1994:
This PR was a mistake. The dense connectivity comes from the concatenation of prior filters, **not** from the _visual representation of the dense connections_. The fact that the "resnet looking" dense architecture works far better for image segmentation and classification is proof of this. 

Sorry for the confusion. I am updating the models on my end, and once I have trained CIFAR 10, 100 and ported the weights from https://github.com/flyyufelix/DenseNet-Keras to Keras properly without usage of the custom Scale layers, I will update file a proper PR. This may take a few days, depending on many factors.

Alejandro-Valdes:
@titu1994 this is super late, and you might not see this. But could you explain why currently the down path is more 'resnet looking' using concatenation of prior filters, as in the left side of the above image, while the upward path has a more 'dense' representation with more concatenations and currently gets models like the right side of the image??

Alejandro-Valdes:
Model representation of encoder

image

Model representation of decoder

image

titu1994:
The simplest and most practical reason is because the weights load properly and give correct validation accuracy on ImageNet valset. 

A more thorough explanation is to think of DenseNets as ResNets which use concatenation instead of summation as their residual. From that perspective, this "ResNet like" connectivity makes more sense. Ofc, DenseNets have a few interesting tricks to prevent naive explosion in parameters and memory consumption. 

My mistake during the initial implementation was to try to imitate the block diagram in their paper rather than the formulation of DenseNets directly.

Alejandro-Valdes:
Thanks I get that, however I dont quite understand why the actual model representation differs so much right now in the downwards and upwards path.

The downward path (encoder) is modeled using resnet connectivity, however the upwards path (decoder) the model isn't modeled like the resnet connectivity, but like the article representation. 

This might give the same results but I still think, that there may be a consistency issue.

PR #144, opened 2017-09-06
Refactor DenseNets

titu1994:
This is a total refactoring of DenseNet model builders. There is now equivalence between this and the PyTorch implementation by the authors of the paper. 

Includes : 
- Pre-built models for ImageNet (ported from https://github.com/flyyufelix/DenseNet-Keras)
- Weights for DenseNetImageNet121 / 161 and 169. **Note** : Requesting someone to port the 201 layer DenseNet weights from the Caffe repo using the repo mentioned above (since it seems caffe has an annoying requirement for the Scale layer) @flyyufelix If possible, could you do that? I can then translate the weights to this repo format.
- DenseNet builder now defaults to CIFAR mode (with no pooling in initial block). Use new flag `subsample_initial_block=True` to use ImageNet mode.
- Major code cleanup so that there is minimal corrections during merge with Keras (if it ever gets merged there)

Due to the major refactors, earlier model weights will probably no longer work. However, try using by_name parameter during weight loading to see if the weights work still (I highly doubt it).

### DIFF
--- a/keras_contrib/applications/densenet.py
+++ b/keras_contrib/applications/densenet.py
@@ -231,7 +231,7 @@ def DenseNet(input_shape=None,
         weights_loaded = False
 
         if (depth == 121) and (nb_dense_block == 4) and (growth_rate == 32) and (nb_filter == 64) and \
-                (bottleneck is True) and (reduction == 0.5) and (subsample_initial_block):
+                (bottleneck is True) and (reduction == 0.5) and (dropout_rate == 0.0) and (subsample_initial_block):
             if include_top:
                 weights_path = get_file('DenseNet-BC-121-32.h5',
                                         DENSENET_121_WEIGHTS_PATH,
@@ -241,12 +241,12 @@ def DenseNet(input_shape=None,
                 weights_path = get_file('DenseNet-BC-121-32-no-top.h5',
                                         DENSENET_121_WEIGHTS_PATH_NO_TOP,
                                         cache_subdir='models',
-                                        md5_hash='8804bcb37da5be4a52dc4e45d4425ba7')
+                                        md5_hash='55e62a6358af8a0af0eedf399b5aea99')
             model.load_weights(weights_path)
             weights_loaded = True
 
         if (depth == 161) and (nb_dense_block == 4) and (growth_rate == 48) and (nb_filter == 96) and \
-                (bottleneck is True) and (reduction == 0.5) and (subsample_initial_block):
+                (bottleneck is True) and (reduction == 0.5) and (dropout_rate == 0.0) and (subsample_initial_block):
             if include_top:
                 weights_path = get_file('DenseNet-BC-161-48.h5',
                                         DENSENET_161_WEIGHTS_PATH,
@@ -256,12 +256,12 @@ def DenseNet(input_shape=None,
                 weights_path = get_file('DenseNet-BC-161-48-no-top.h5',
                                         DENSENET_161_WEIGHTS_PATH_NO_TOP,
                                         cache_subdir='models',
-                                        md5_hash='d38903b8732fe238c91dac7859271f26')
+                                        md5_hash='1a9476b79f6b7673acaa2769e6427b92')
             model.load_weights(weights_path)
             weights_loaded = True
 
         if (depth == 169) and (nb_dense_block == 4) and (growth_rate == 32) and (nb_filter == 64) and \
-                (bottleneck is True) and (reduction == 0.5) and (subsample_initial_block):
+                (bottleneck is True) and (reduction == 0.5) and (dropout_rate == 0.0) and (subsample_initial_block):
             if include_top:
                 weights_path = get_file('DenseNet-BC-169-32.h5',
                                         DENSENET_169_WEIGHTS_PATH,
@@ -271,7 +271,7 @@ def DenseNet(input_shape=None,
                 weights_path = get_file('DenseNet-BC-169-32-no-top.h5',
                                         DENSENET_169_WEIGHTS_PATH_NO_TOP,
                                         cache_subdir='models',
-                                        md5_hash='a664d78a30ddd217dd38c0bb8d258461')
+                                        md5_hash='89c19e8276cfd10585d5fadc1df6859e')
             model.load_weights(weights_path)
             weights_loaded = True
 
--- a/keras_contrib/applications/densenet.py
+++ b/keras_contrib/applications/densenet.py
@@ -92,7 +92,7 @@ def preprocess_input(x, data_format=None):
     # Returns
         Preprocessed tensor.
     """
-    x = _preprocess_input(x, data_format=None)
+    x = _preprocess_input(x, data_format=data_format)
     x *= 0.017  # scale values
     return x
 
--- a/keras_contrib/applications/densenet.py
+++ b/keras_contrib/applications/densenet.py
@@ -1,5 +1,38 @@
 # -*- coding: utf-8 -*-
-'''DenseNet models for Keras.
+'''DenseNet and DenseNet-FCN models for Keras.
+
+DenseNet is a network architecture where each layer is directly connected
+to every other layer in a feed-forward fashion (within each dense block).
+For each layer, the feature maps of all preceding layers are treated as
+separate inputs whereas its own feature maps are passed on as inputs to
+all subsequent layers. This connectivity pattern yields state-of-the-art
+accuracies on CIFAR10/100 (with or without data augmentation) and SVHN.
+On the large scale ILSVRC 2012 (ImageNet) dataset, DenseNet achieves a
+similar accuracy as ResNet, but using less than half the amount of
+parameters and roughly half the number of FLOPs.
+
+DenseNets support any input image size of 32x32 or greater, and are thus
+suited for CIFAR-10 or CIFAR-100 datasets. There are two types of DenseNets,
+one suited for smaller images (DenseNet) and one suited for ImageNet,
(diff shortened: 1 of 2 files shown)


titu1994:
@athundt Could you review it when you have time? I will be busy for the rest of the week, so I may respond sporadically.

keras_contrib/applications/densenet.py
```diff
@@ -1,5 +1,38 @@
...
+|   DenseNet-161    |    22.20 %            |         -   %         |     28.9     |
+------------------------------------------------------------------------------------
+
+DenseNets can be extended to image segmentation tasks as described in the
+paper "The One Hundred Layers Tiramisu: Fully Convolutional DenseNets for
+Semantic Segmentation". Here, the dense blocks are arranged and concatenated
+with long skip connections for state of the art performance on CamVid dataset.
+
 # Reference
 - [Densely Connected Convolutional Networks](https://arxiv.org/pdf/1608.06993.pdf)
```
ahundt: Add:

```
This implementation is based on the following reference code:
 - https://github.com/gpleiss/efficient_densenet_pytorch
 - https://github.com/liuzhuang13/DenseNet
```

keras_contrib/applications/densenet.py
```diff
@@ -442,38 +725,58 @@ def __create_dense_net(nb_classes, img_input, include_top, depth=40, nb_dense_bl
         bottleneck: add bottleneck blocks
         reduction: reduction factor of transition blocks. Note : reduction value is inverted to compute compression
         dropout_rate: dropout rate
-        weight_decay: weight decay
+        weight_decay: weight decay rate
+        subsample_initial_block: Set to True to subsample the initial convolution and
```
ahundt: Add something like `Typically True for imagenet pretrained weights, False for cifar pretrained weights`.

keras_contrib/applications/densenet.py
```diff
@@ -130,47 +244,69 @@ def DenseNet(input_shape=None, depth=40, nb_dense_block=3, growth_rate=12, nb_fi
...
-                                  '(`image_data_format="channels_first"`). '
-                                  'For best performance, set '
-                                  '`image_data_format="channels_last"` in '
-                                  'your Keras config '
-                                  'at ~/.keras/keras.json.')
-                    convert_all_kernels_in_model(model)
+    if weights == 'imagenet':
+        weights_loaded = False
+
+        if (depth == 121) and (nb_dense_block == 4) and (growth_rate == 32) and (nb_filter == 64) and \
```
ahundt: add check that epsilon matches

ahundt:
Oh also be sure to use @ahundt, that's my actual username

keras_contrib/applications/densenet.py
```diff
@@ -11,89 +44,169 @@
...
+             reduction=0.0,
+             dropout_rate=0.0,
+             weight_decay=1e-4,
+             subsample_initial_block=False,
+             include_top=True,
+             weights=None,
+             input_tensor=None,
+             pooling=None,
+             classes=10,
+             activation='softmax'):
```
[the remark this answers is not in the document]
titu1994: Epsilon is not a hyperparameter. The original code defaults to it for some reason, and the only reason I was forced to use it as well was the fact that the weights were trained as such using this default.

In actuality, epsilon is too small of a factor in batch normalization to affect training severely.

titu1994:
@ahundt Sorry about the username mismatch. Made all corrections as requested, other than making epsilon a parameter. 

The value for epsilon is not significant during training, and I would have liked to keep it the Keras default. However, the weights provided via Caffe require a different value. It does not affect training as much as it affects the inference using pre-trained weights.

Epsilon in itself has very little impact on BatchNormalization during training.

keras_contrib/applications/densenet.py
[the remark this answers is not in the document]
ahundt: that's reasonable

ahundt:
@farizrahman4u @the-moliver could one of you handle the merge step?

keras_contrib/applications/densenet.py
ahundt: not necessary as per https://github.com/farizrahman4u/keras-contrib/pull/144#discussion_r137431139

keras_contrib/applications/densenet.py
```diff
@@ -297,95 +416,182 @@ def DenseNetFCN(input_shape, nb_dense_block=5, growth_rate=16, nb_layers_per_blo
...
+                    include_top=include_top, weights=weights, input_tensor=input_tensor,
+                    pooling=pooling, classes=classes, activation=activation)
+
+
+def DenseNetImageNet161(input_shape=None,
+                        bottleneck=True,
+                        reduction=0.5,
+                        dropout_rate=0.0,
+                        weight_decay=1e-4,
+                        include_top=True,
```
ahundt: should all the include_top instances be updated to require_flatten?

titu1994: The name remains include top in Keras main, just the internal call to _obtain_image_size has require_flatten.

keras_contrib/applications/densenet.py
```diff
@@ -106,7 +202,7 @@ def DenseNet(input_shape=None, depth=40, nb_dense_block=3, growth_rate=12, nb_fi
                                       default_size=32,
                                       min_size=8,
```
ahundt: does min_size and default_size change for imagenet vs cifar weights?

titu1994: I was able to load both sets of weights for all 3 models with no alterations to this, so I didn't set specific min size and default size. Seems as long as an input size is provided, there is no issue with this.

ahundt:
@titu1994 I tried downloading the pretrained weights and the md5sum may not match the one provided in the code?

```
Downloading data from https://github.com/titu1994/DenseNet/releases/download/v3.0/DenseNet-BC-121-32-no-top.h5
29671424/33199896 [=========================>....] - ETA: 0sWeights for the model were loaded successfully
A local file was found, but it seems to be incomplete or outdated because the md5 file hash does not match the original value of 8804bcb37da5be4a52dc4e45d4425ba7 so we will re-download the data.
Downloading data from https://github.com/titu1994/DenseNet/releases/download/v3.0/DenseNet-BC-121-32-no-top.h5
29351936/33199896 [=========================>....] - ETA: 0sWeights for the model were loaded successfully
± md5sum ~/.keras/models/DenseNet-BC-121-32-no-top.h5
55e62a6358af8a0af0eedf399b5aea99  /home/ahundt/.keras/models/DenseNet-BC-121-32-no-top.h5
```

titu1994:
Sorry, I forgot to update the md5 after I updated the weights. I'll be able to do that in a 3-4 hours.

titu1994:
@ahundt I was wondering, now that DenseNets have ImageNet weights, wouldnt it be more suitable to add it to Keras proper itself? 

I would likely have to remove some parts, like the entire FCN part, which will remain in Keras contrib.

ahundt:
I looked again and I think you accidentally added the dropout check back in in your recent commit.

Regarding a PR to keras proper, I think a PR to keras might be accepted, though some of the changes he will require would make my use case more difficult, (not having the `_create_*` functions).

Also I'm going to add a quick PR to your master repo to add in `with K.name_scope()` scopes so it is easier to look at the network layout.

titu1994:
Actually, I realised that the dropout check is needed for weight loading. 

My code checks whether dropout rate is 0 to decide if it should add a dropout layer or no. While loading weights, if there exists a dropout layer that does not exist in the weights, it causes problems with weight loading unless "by_name" is set to true. 

Ues the loss of the create function is lamentable, but I was thinking I'll keep DenseNet here as well. Until DenseNet FCN weights are obtained, ported and then merged into Keras. Even then, Keras proper will probably not need the extra Subpixel later, or the additional backend functions, so perhaps FCN will always remain here. 

As to your addition of name scopes, unless it can be done without a ton of if else checks on the backend of Keras, please do no add it. I opt for simplicity and readability of the code, not the model it produces. A bunch of backend checks just destroys that. This was the reason that I did not add TF name scopes in any repo till date.

ahundt:
> As to your addition of name scopes, unless it can be done without a ton of if else checks on the backend of Keras, please do no add it. I opt for simplicity and readability of the code, not the model it produces. A bunch of backend checks just destroys that. This was the reason that I did not add TF name scopes in any repo till date.

No if/else statements, just ~4-5 new lines and some added white space for indentation.

ahundt:
> My code checks whether dropout rate is 0 to decide if it should add a dropout layer or no. While loading weights, if there exists a dropout layer that does not exist in the weights, it causes problems with weight loading unless "by_name" is set to true.

You have a fair point, but perhaps we should either (1) always create dropout since there should be no runtime cost and update the weights files again, (2) set "by_name" to true? or (3) leave the weights + dropout as-is.

titu1994:
@ahundt Lets go with option 2. I realised that people who perform fine tuning may wish to change dropout value if they are not performing data augmentation, and the check would cause weight loading to fail. 

Also, merged your PR. While I have no problems keeping it in contrib, I think if/when the DenseNet code is merged onto Keras, fchollet will request removal of the name scopes. No other application in Keras proper has them (even Inception variants, which arguably would _really_ benefit from it). We can cross that bridge when we have to face it I guess.

titu1994:
@farizrahman4u Could you merge this please?

ahundt:
> @farizrahman4u Could you merge this please?

+1

titu1994:
@kemaswill @farizrahman4u

ahundt:
@tboquet are you possibly able to take a look at this?

PR #145, opened 2017-09-06
densenet updates + pretrained weights

ahundt:
Incorporates changes from https://github.com/titu1994/DenseNet from @titu1994 including imagenet pretrained weights.

Before merging, I also wanted to discuss which of these differences are correct updates, which aspects from the existing code should we keep?

- dropout removed from bottleneck (this is still in upstream)
- bottleneck lines removed 
```
   if bottleneck:
        nb_layers = [int(layer // 2) for layer in nb_layers]
```
- regularizers changed (add epsilon, remove alpha beta regularizers)
- weight decay parameter not passed through in some cases, added in others
- initial convolution goes from 3x3 to 7x7
- should there be a conv-bn-relu vs bn-relu-conv option? (latter improved resnetv2 results)

### DIFF
--- a/keras_contrib/applications/densenet.py
+++ b/keras_contrib/applications/densenet.py
@@ -13,7 +13,7 @@ import warnings
 from keras.models import Model
 from keras.layers.core import Dense, Dropout, Activation, Reshape
 from keras.layers.convolutional import Conv2D, Conv2DTranspose, UpSampling2D
-from keras.layers.pooling import AveragePooling2D
+from keras.layers.pooling import AveragePooling2D, MaxPooling2D
 from keras.layers.pooling import GlobalAveragePooling2D
 from keras.layers import Input
 from keras.layers.merge import concatenate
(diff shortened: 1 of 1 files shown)


titu1994:
@ahundt I already added a PR for this?

titu1994:
In any case, I will answer the points when comparing this model to the PyTorch implementation suggested by the authors :

> dropout removed from bottleneck (this is still in upstream)

There is supposed to be only 1 dropout in the dense block, which is at the main conv. The extra dropout at the bottleneck is not present on the PyTorch implementation.

> if bottleneck:
        nb_layers = [int(layer // 2) for layer in nb_layers]

This accounts for the missing conv layers in a bottleneck model. This was an incorrect assumption on my part, that due to the addition of bottleneck conv, the layer would have 2 convs instead of 1 and therefore half the number of layers to account for that. It was incorrect. The weights ported require this line to be removed.

> regularizers changed (add epsilon, remove alpha beta regularizers)

There is no regularizers for BatchNorm. The epsilon change is crucial for the weights loaded to have correct prediction strength. Without it, the correct predictions are made, but the strength of the predictions is on the order of 1-2 % for the strongest class, with it it jumps to the correct 90% + for cheetah image (test it in my repo)

>  weight decay parameter not passed through in some cases, added in others

Where is the weight decay parameter not passed? I added the weight decay to all 4 block functions.

> initial convolution goes from 3x3 to 7x7

This is correct. The Inception mode DenseNet has a 7x7 stride 2 convolution. The CIFAR mode densenet uses a 3x3 stride 1 convolution.

> should there be a conv-bn-relu vs bn-relu-conv option? (latter improved resnetv2 results)

There are already too many parameters to modify. I would not add this. If it is really in the interest of someone to do this, they can copy the entire script and modify as needed. The latter may have improved ResNetV2, but it is not guaranteed to improve a DenseNet. 

Furthermore, there is no literature saying one is better than the other _always_, just empirical results from the ResNet paper on a specific model. Whether it transfers to other models has not been studied yet.

ahundt:
All excellent points, thanks! I'll take a look at your PR (I had not seen it) and close this.

titu1994:
@ahundt No problem, and they were important questions. Glad I was able to get it done asap.

PR #146, opened 2017-09-07
Fix for Python 3 in conll dataset loading

vilmar-hillow:


ahundt:
have you tested this on python2?

vilmar-hillow:
Yes, tested both on 2 and 3

PR #147, opened 2017-09-08
wide_resnet.py keras 2 update

ahundt:


titu1994:
I was wondering, shouldn't we replace the average pooling 8x8 by a GlobalAveragePooling2D? Its cleaner that way, and should continue to work properly. We can also remove the flatten then.

ahundt:
I'm converting the code to work for segmentation problems as well, so in my case the change you describe would unfortunately be counterproductive.

ahundt:
@titu1994 could you give this a quick review? I've been using it for a few weeks without issue now.

titu1994:
You could add the pooling argument like the rest of the models and select what kind of pooling should occur. 

I think DenseNet implements this already so you could add it from there. Otherwise everything else is fine.

ahundt:
switched to global average pooling as you suggested

ahundt:
This change seems pretty trivial at this point and the error is unrelated, so I'll merge this.

PR #151, opened 2017-09-22
Fix normalization tests

titu1994:
@ahundt This should fix the tests. If it doesn't I will look into it over the weekend.


titu1994:
The test failure was due to some random could not import problems. The tests passed locally. Let's hope it completes it properly this time

titu1994:
I have no clue as to why the tests are failing due to "cannot import from backend" errors. @ahundt @farizrahman4u Any solutions? The batchnorm tests as they are all passing locally.

titu1994:
Update: Just ran the tests folder by folder, now it says there are errors in the ConvolutionalAware test. I have no idea what that is, so I can't fix it.

PR #121, opened 2017-07-17
Adding DICE to loss functions

kmader:
adding dice / dice_loss to the loss functions

PR #152, opened 2017-09-24
Add jaccard distance loss

wassname:
Adds jaccard distance loss. 

Why do we want this? It's better than categorical cross entropy when you have unbalanced data within a sample e.g. for pixel-wise segmentation. For this reason it (and the similar dice loss) has been been [used in kaggle competition with great results](https://github.com/ZFTurbo/KAGGLE_CERVICAL_CANCER_2017).

I modified this version to converge from greater than zero to zero because it looks nicer and matches other losses. I've also smoothed it to have a smooth gradient even as it approaches zero loss (otherwise it may stop short of max accuracy like dice loss).

Why not dice loss? I have found this to be similar in application but superior in results to dice loss #121 because dice loss has a vanishing gradient as it approached zero loss. So if you merge this you might want to consider closing the #121 without merging.


wassname:
pep8 and tests pass on my system for tensorflow and py3.5. They fail on travis due to existing errors in keras-contrib.

ahundt:
This needs a reference to the original paper.

wassname:
Alright sounds good, I'll try to do that in the next few weeks

ahundt:
I also forgot one important one:

- add a usage example to the examples folder that illustrates the advantage of this loss function

examples/pascal_voc_jaccard_loss.py
```diff
@@ -0,0 +1,84 @@
+'''
+Trains a DenseNet-40-12 model on the CIFAR-10 Dataset.
```
ahundt: shouldn't this file be called something like cifar10_jaccard_loss.py?

wassname: Ops I need to change that docstring, it has to use pascal voc because it's showing how jaccard loss is good for segmention.

examples/pascal_voc_jaccard_loss.py
```diff
@@ -0,0 +1,129 @@
...
+seed = 1
+image_generator = image_datagen.flow_from_directory(
+    pascal_folder,
+    target_size=(img_rows, img_cols),
+    class_mode=None,
+    classes=["img"],
+    seed=seed)
+mask_generator = image_datagen.flow_from_directory(
+    pascal_folder,
+    target_size=(img_rows, img_cols),
```
ahundt: resizing the labels this way will change the label values, and training won't work correctly. Consider merging SegDataGenerator from https://github.com/aurora95/Keras-FCN/blob/master/utils/SegDataGenerator.py#L291. A usage example is available at https://github.com/aurora95/Keras-FCN/blob/master/train.py#L117

wassname: Oh shit, you're right. Thanks for catching that. I'll do that or a basic decimation (I only have 2GB gpu).

ahundt:
@wassname I made a PR for SegDataGenerator in case you're up for testing it out so we can finish up this PR https://github.com/farizrahman4u/keras-contrib/pull/195

wassname:
Great stuff! I'll give it a test

wassname:
The augmentation works but it isn't converging on pascal VOC data, probably due to [this issue](https://github.com/farizrahman4u/keras-contrib/issues/63). 

I'm going to close this issue since I don't have time for the amount of work you've asked me to do to add this loss function. Making an example script, fixing all the things associated with it, training, documentation, and a unit test is going to take more time than it's worth for a simple loss function I'm afraid.

Anyone else is welcome to continue this, and anyone who want's to use the function can refer to [this gist](https://gist.github.com/wassname/f1452b748efcbeb4cb9b1d059dce6f96).

ahundt:
@wassname Sorry! I didn't mean to imply that the performance on Pascal VOC would hold up this PR, I was just explaining what was going on. Let me give it one last look and this can probably be merged.

Do you have the changes you tested with SegDataGenerator somewhere? I'd like to take a look.

Also, I don't know how others feel about it, but I'd love to have that loss visualization included. There should be more examples like that which help people understand the functions.

ahundt:
@wassname Just a tiny pep8 issue, and if you wouldn't mind adding a note about the training performance.

```
__________________ PEP8-check(ignoring E501 \ E402 \ E731 \) ___________________

/home/travis/build/farizrahman4u/keras-contrib/keras_contrib/losses/jaccard.py:18:195: W291 trailing whitespace

    - Csurka, Gabriela & Larlus, Diane & Perronnin, Florent. (2013). What is a good evaluation measure for semantic segmentation?. IEEE Trans. Pattern Anal. Mach. Intell.. 26. . 10.5244/C.27.32.

                                                                                                                                                                                                  ^


```

wassname:
Ok great! I've pushed the version that uses SegDataGenerator. 

I'll check it for PEP8 and make that note in the next few days.

ahundt:
@wassname I made a separate PR #201 which separates the SegDataGenerator stuff out and includes the pep8 fix, it just took a moment and I improved the doc wording a bit. If #201 passes travis I'll merge it. 

When I have a chance I'll also finish up the SegDataGenerator PR and probably use your example as a starting point, but that will be a few days. Thanks!

ahundt:
I'll close this again in favor of #201 to de-clutter

wassname:
Sounds good, thanks!

HasnainRaz:
Hey, I was interested in trying out jaccard loss for my problems, but I'm not sure if the implementation is correct. The absolute symbol in the jaccard loss is the cardinality of the set, not the absolute value. Can anyone explain the math behind why the absolute value is used? Am I missing something here?

ahundt:
@HasnainRaz I can't revisit this in detail at the moment myself, so take my comment with a grain of salt, but I believe for certain problem types the abs isn't a problem. If you have a more general implementation could you consider submitting a PR with a test that fails but should pass with the current code + a fix?

PR #155, opened 2017-09-25
setup.py update to match keras

ahundt:


ahundt:
This worked on my local tests, and others don't seem available for review so I'm going to merge it since it is very minor and can easily be reverted if there is a problem.

PR #156, opened 2017-09-25
recurrent.py _time_distributed_dense doesn't exist

ahundt:
upstream keras doesn't have the _time_distributed_dense symbol anymore


ahundt:
TF and pep8 at least pass again with this, so I'm merging it. theano still has something going on...

Issue #157, opened 2017-09-25
fix travis - h5py model saving issue


ahundt:
@the-moliver @bstriner it looks like the current remaining travis failure might be due to the hdf5 saving function, within files it appears you have authored.

https://github.com/farizrahman4u/keras-contrib/blob/master/tests/keras_contrib/utils/save_load_utils_test.py#L36
Failure: https://travis-ci.org/farizrahman4u/keras-contrib/builds/279618487?utm_source=github_status&utm_medium=notification

Oh it looks like there is some id that occurs twice, and that shouldn't work. Should the test be changed or the code here?

```
________________________ test_save_and_load_all_weights ________________________
[gw1] linux2 -- Python 2.7.14 /home/travis/miniconda/envs/test-environment/bin/python
    def test_save_and_load_all_weights():
        '''
        Test save_all_weights and load_all_weights. Save and load optimizer and model weights but not configuration.
        '''
    0m
        def make_model():
            _x = Input((10,))
            _y = Dense(10)(_x)
            _m = Model(_x, _y)
            _m.compile('adam', 'mean_squared_error')
            _m._make_train_function()
            return _m
    
        # make a model
        m1 = make_model()
        # set weights
        w1 = m1.layers[1].kernel  # dense layer
        w1value = K.get_value(w1)
        w1value[0, 0:4] = [1, 3, 3, 7]
        K.set_value(w1, w1value)
        # set optimizer weights
        ow1 = m1.optimizer.weights[3]  # momentum weights
        ow1value = K.get_value(ow1)
        ow1value[0, 0:3] = [4, 2, 0]
        K.set_value(ow1, ow1value)
        # save all weights
>       save_all_weights(m1, 'model.h5')
tests/keras_contrib/utils/save_load_utils_test.py:37: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
keras_contrib/utils/save_load_utils.py:74: in save_all_weights
    dtype=val.dtype)
../../../miniconda/envs/test-environment/lib/python2.7/site-packages/h5py/_hl/group.py:111: in create_dataset
    self[name] = dset
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
[...]

PR #161, opened 2017-10-10
save_load_utils_test.py removes 'model.h5'

ahundt:
https://github.com/farizrahman4u/keras-contrib/issues/157

If the build passes with this change I'll just merge it.

### DIFF
--- a/tests/keras_contrib/utils/save_load_utils_test.py
+++ b/tests/keras_contrib/utils/save_load_utils_test.py
@@ -1,4 +1,5 @@
 import pytest
+import os
 from keras import backend as K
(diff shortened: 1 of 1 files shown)


titu1994:
Hopefully this fixes the buggy tests.

ahundt:
Apparently it does not fix the bug... but it doesn't break anything either.

ahundt:
Oh it looks like there is some id that occurs twice, and that shouldn't work. @bstriner should the test be changed or the code here?

ahundt:
this at least removes files as the tests should... so I'm merging it

PR #167, opened 2017-10-17
TimeDistributed ResNet + test

ahundt:
Adds a unit test for resnet models and enables creation of a TimeDistributed ResNet


keras_contrib/applications/resnet.py
```diff
@@ -12,11 +12,16 @@
...
+Implementation adapted from: github.com/raghakot/keras-resnet
+
+TimeDistributed adapted from: github.com/voletiv/keras-resnet
 """
 from __future__ import division
 
 import six
+import warnings
+from keras.utils.data_utils import get_file
+from keras.utils import layer_utils
```
voletiv: 22-24 (import warnings, get_file, layer_utils) are unnecessary I think, because they are not used anywhere

keras_contrib/applications/resnet.py
```diff
@@ -338,15 +481,19 @@ def ResNet(input_shape=None, classes=10, block='bottleneck', residual_unit='v2',
...
-    if len(input_shape) != 3:
-        raise Exception("Input shape should be a tuple (nb_channels, nb_rows, nb_cols)")
+    input_shape = obtain_input_shape(input_shape=input_shape, require_flatten=include_top,
+                                     time_distributed=time_distributed)
+
+    _handle_dim_ordering(time_distributed=time_distributed, verbose=verbose)
+
+    if time_distributed:
+        if len(input_shape) != 4:
+            raise Exception("Input shape should be a tuple (nb_time_steps, nb_rows, nb_cols, nb_channels)")
```
voletiv: Since we are leaving out permutation of dims, maybe we should print an image_data_format-specific message here?

```
if time_distributed:
    if len(input_shape) != 4:
        if K.image_data_format() == 'channels_last':
            raise Exception("channels_last: Input shape should be a tuple (nb_time_steps, nb_rows, nb_cols, nb_channels)")
        else:
            raise Exception("channels_first: Input shape should be a tuple (nb_time_steps, nb_channels, nb_rows, nb_cols)")
else:
    if len(input_shape) != 3:
        if K.image_data_format() == 'channels_last':
            raise Exception("channels_last: Input shape should be a tuple (nb_rows, nb_cols, nb_channels)")
        else:
            raise Exception("channels_first: Input shape should be a tuple (nb_channels, nb_rows, nb_cols)")


```

keras_contrib/applications/resnet.py
```diff
@@ -366,20 +513,22 @@ def ResNet(input_shape=None, classes=10, block='bottleneck', residual_unit='v2',
...
-    if K.image_data_format() == 'channels_first':
-        input_shape = (input_shape[1], input_shape[2], input_shape[0])
-    # Determine proper input shape
-    input_shape = _obtain_input_shape(input_shape,
-                                      default_size=32,
-                                      min_size=8,
-                                      data_format=K.image_data_format(),
-                                      require_flatten=include_top)
+    # Determine proper input shape again
+    input_shape = obtain_input_shape(input_shape=input_shape, require_flatten=include_top,
```
voletiv: No need to "obtain_input_shape" again. That was included in the original file because there was a permutation of dim ordering before this, which you deleted.

keras_contrib/applications/resnet.py
```diff
@@ -328,6 +468,9 @@ def ResNet(input_shape=None, classes=10, block='bottleneck', residual_unit='v2',
         top: Defines final layers to evaluate based on a specific problem type. Options are
             'classification' for ImageNet style problems, 'segmentation' for problems like
             the Pascal VOC dataset, and None to exclude these layers entirely.
+        time_distributed: Boolean that indicates whether the ResNet is to be repeated over a
+            time axis, to make a TimeDistributed ResNet
+        verbose: Boolean that indicates whether to print layer names as they are built
```
voletiv: Let's add a comment on "weights" argument before "verbose"?

ahundt:
@voletiv ready for another look :-) thanks for the review!

ahundt:
ugh the unit test takes too long, but it only constructs several variations on resnet... I guess I'll just drop some of the larger network tests?

voletiv:
@ahundt Might as well squash these minor commits?
(I didn't want to be seem too picky to mention stage=stage..!)

Also, maybe keep a prediction in test_resnet.py? As I mentioned in #192, mere compilation is not an effective metric for correct working.

ahundt:
so close! however, the predict line made the test too slow so it failed

voletiv:
Predict only on the tiniest input_dim, (8, 8, 1)

keras_contrib/applications/resnet.py
```diff
@@ -490,10 +492,16 @@ def ResNet(input_shape=None, classes=10, block='bottleneck', residual_unit='v2',
 
     if time_distributed:
         if len(input_shape) != 4:
-            raise Exception("Input shape should be a tuple (nb_time_steps, nb_rows, nb_cols, nb_channels)")
+            if K.image_data_format() == 'channels_last':
```
voletiv: Sorry for the late observation, this check for input_shape dimensions needs to be before "# Determine proper input shape".

voletiv:
@ahundt Here's how I make your life much simpler.

Turns out all we have to do to let TimeDistributed(Resnet18()) work is set the "learning_phase" backed variable in Keras.

```
import numpy as np

from keras.layers import Input, TimeDistributed
from keras import backend as K
from resnet import *

# In case of training the resnet
K.set_learning_phase(True)
# # In case of predicting using trained resnet
# K.set_learning_phase(False)

input_shape = (8, 8, 1,)
TIME_STEPS = 3
time_distributed_input_shape = (TIME_STEPS,) + input_shape
i = Input(shape=time_distributed_input_shape)
time_distributed_resnet = TimeDistributed(ResNet18(input_shape=input_shape, classes=10))
o = time_distributed_resnet(i)
model = Model(inputs=i, outputs=o)
model.compile(optimizer="sgd", loss="categorical_crossentropy")

a = np.random.random((2,) + time_distributed_input_shape)
b = model.predict(a)

```

Turns out `K.learning_phase()` is a parameter required by the `BatchNormalization` layers for this to work. (I am embarrassed.)

By the way, setting `K.set_learning_phase` to `True` or `False` does not matter to run `model.predict(a)`.

ahundt:
hmm but if you want to train on TimeDistributed data what do you do?

ahundt:
All the tests that pass on master appear to pass here now.

voletiv:
@ahundt I'm saying my PR is not required. @titu1994 's naive way does work, provided the `learning_phase` variable is set.

Set `K.set_learning_phase(True)` when training needs to be done.

ahundt:
I got that part, I think my question is slightly different. 

But what about something like the following from a [machine learning mastery tutorial](https://machinelearningmastery.com/timedistributed-layer-for-long-short-term-memory-networks-in-python/)?

```
from numpy import array
from keras.models import Sequential
from keras.layers import Dense
from keras.layers import TimeDistributed
from keras.layers import LSTM
# prepare sequence
length = 5
seq = array([i/float(length) for i in range(length)])
X = seq.reshape(1, length, 1)
y = seq.reshape(1, length, 1)
# define LSTM configuration
n_neurons = length
n_batch = 1
n_epoch = 1000
# create LSTM
model = Sequential()
model.add(LSTM(n_neurons, input_shape=(length, 1), return_sequences=True))
model.add(TimeDistributed(Dense(1)))
model.compile(loss='mean_squared_error', optimizer='adam')
print(model.summary())
# train LSTM
model.fit(X, y, epochs=n_epoch, batch_size=n_batch, verbose=2)
# evaluate
result = model.predict(X, batch_size=n_batch, verbose=0)
for value in result[0,:,0]:
	print('%.1f' % value)
```

Doesn't training have to be true in this case, and a TimeDistributed is instantiated, so wouldn't that model you described not train properly in the case that you want to both train and predict? (not just predict)

voletiv:
learning_phase has to be True in case of training.

Regarding the tutorial, the tutorial wants a TimeDistributed output to the LSTM layer. In our case, we want a TimeDistributed ResNet before an LSTM (say). Something like this:
http://preon.iiit.ac.in/~vikram_voleti/posts/20171117/LSTM_Assessor.png
(the CNN being a ResNet in our case)

So, wrapping TimeDistributed over ResNet achieves this, as long as learning_phase is set. (as far as I know, I haven't tried training with the naive implementation.)

ahundt:
Oh I see, initially it is undefined, and if you set it to anything at all then it is created properly. So strange!

ahundt:
Any chance you might be willing to re-update it to remove TimeDistributed? I fixed a bunch of other bugs anyway so quite a few of the changes should be kept.

voletiv:
Yup, I'll do that.

ahundt:
I'm merging this version for now because it fixes bugs in the existing version and while there is a bit of extraneous code, it works compared to the current master version, which does not. I'd appreciate a PR removing time_distributed when you're able! Thanks!

voletiv:
I've sent a pull request to ahundt:resnet-test, please check

titu1994:
Is there any update to this ? I've been considering building the LM-ResNets and it would be useful to have something to base it on. 

Also, this script base is extremely complex in my opinion. It uses method wrappers, layer name generators and incorporates a lot of logical overhead for simple models like ResNets.  Can this be refactored to be cleaner ?

ahundt:
I use this code on a daily basis, it works pretty well for me as far as I can tell. I just haven't gotten around to merging this particular branch with master again to see if the tests pass.

ahundt:
If you'd like to simplify it I'd be delighted to merge that here then we can finally merge the PR once the unit tests pass. I just haven't had the motivation to refactor.

PR #169, opened 2017-10-17
remove Deconvoluion3D, add missing backend functions

ahundt:
Hopefully this should also fix the unit test failures.
Deconv3D is available as Conv3DTranspose in Keras https://github.com/fchollet/keras/pull/7161 so it is no longer required in keras-contrib.


titu1994:
Seems there's a similar problem with the conv2d version of this function as well.

ahundt:
Note that revision 80d698c contains code from https://github.com/fchollet/keras/blob/1bbfdb6174c2833c7aca8261a3f97bc5235bc66b/keras/backend/tensorflow_backend.py

ahundt:
tests pass again so I'm merging this.

PR #173, opened 2017-10-24
Swish activation function

abhaikollara:
Paper can be found [here](https://arxiv.org/abs/1710.05941)


ahundt:
Thanks, should this also be wrapped in a layer or something? Just be sure to match how keras implements their activations upstream. Also there will need to be a unit test. A bunch of other people implemented swish and made PRs to keras itself, so there might be something already done over there you could utilize.

ahundt:
closing in favor of #181

PR #174, opened 2017-10-25
Fixed and extended Dead ReLU detector callback

infelane:
Initially 'DeadReLUDetectorCallback was only working for Dense layers and the percentage of dead relu neurons was incorrect.

Corrected the percentage calculations
Extended for all layers with a ReLU activation (tested for conv2d and dense)
Tested on python 3.5 TF and Theano

### DIFF
--- a/keras_contrib/callbacks/dead_relu_detector.py
+++ b/keras_contrib/callbacks/dead_relu_detector.py
@@ -20,7 +20,7 @@ class DeadReluDetector(Callback):
             False means the warning message is printed.
     """
 
-    def __init__(self, x_train, verbose=False, bool_warning = False):
+    def __init__(self, x_train, verbose=False, bool_warning=False):
         super(DeadReluDetector, self).__init__()
         self.x_train = x_train
         self.verbose = verbose
@@ -87,9 +87,8 @@ class DeadReluDetector(Callback):
             if (self.verbose and dead_neurons > 0) or dead_neurons_share > self.dead_neurons_share_threshold:
                 str_warning = 'Layer {} (#{}) has {} dead neurons ({:.2%})!'.format(layer_name, layer_index,
                                                                                     dead_neurons, dead_neurons_share)
-                
+
                 if self.bool_warning:
                     warnings.warn(str_warning, RuntimeWarning)
                 else:
                     print(str_warning)
-
--- a/keras_contrib/callbacks/dead_relu_detector.py
+++ b/keras_contrib/callbacks/dead_relu_detector.py
@@ -2,7 +2,6 @@ import numpy as np
 import warnings
 
 from keras.callbacks import Callback
-from keras.layers import Dense
 from keras import backend as K
 
 
@@ -17,6 +16,7 @@ class DeadReluDetector(Callback):
             False means that only significant number of dead neurons (10% or more)
             triggers warning
     """
+
     def __init__(self, x_train, verbose=False):
         super(DeadReluDetector, self).__init__()
         self.x_train = x_train
@@ -25,7 +25,11 @@ class DeadReluDetector(Callback):
 
     @staticmethod
     def is_relu_layer(layer):
-        return isinstance(layer, Dense) and layer.get_config()['activation'] == 'relu'
+        # Should work for all layers with relu activation. Tested for Dense and Conv2D
+        if 'activation' in layer.get_config():
+            return layer.get_config()['activation'] == 'relu'
+        else:
+            return False
 
     def get_relu_activations(self):
         model_input = self.model.input
@@ -44,17 +48,43 @@ class DeadReluDetector(Callback):
         layer_outputs = [func(list_inputs)[0] for func in funcs]
         for layer_index, layer_activations in enumerate(layer_outputs):
             if self.is_relu_layer(self.model.layers[layer_index]):
-                yield [layer_index, layer_activations]
+                layer_name = self.model.layers[layer_index].name
+                # layer_weight is a list [W] (+ [b])
+                layer_weight = self.model.layers[layer_index].get_weights()
+                # with kernel and bias, the weights are saved as a list [W, b]. If only weights, it is [W]
+                assert type(layer_weight) == list
+                layer_weight_shape = np.shape(layer_weight[0])
+                yield [layer_index, layer_activations, layer_name, layer_weight_shape]
 
     def on_epoch_end(self, epoch, logs={}):
         for relu_activation in self.get_relu_activations():
-            layer_index, activation_values = relu_activation
-            total_neurons = activation_values.shape[-1]
-            dead_neurons = np.sum(activation_values == 0)
-            dead_neurons_share = dead_neurons / total_neurons
+            layer_index, activation_values, layer_name, layer_weight_shape = relu_activation
+
+            shape_act = activation_values.shape
+
+            weight_len = len(layer_weight_shape)
+            act_len = len(shape_act)
+
+            # should work for both Conv and Flat
+            if K.backend() == 'tensorflow':
+                # features in last axis
+                axis_filter = -1
+            elif K.backend() == 'theano':
+                # features before the convolution axis, for weight_len the input and output have to be subtracted
+                axis_filter = -1 - (weight_len - 2)
+            else:
+                raise ValueError('Unknown backend: {}'.format(K.backend()))
+
+            total_featuremaps = shape_act[axis_filter]
+
+            axis = tuple(
+                i for i in range(act_len) if (i != axis_filter) and (i != (len(shape_act) + axis_filter)))
+
+            dead_neurons = np.sum(np.sum(activation_values, axis=axis) == 0)
+
+            dead_neurons_share = dead_neurons / total_featuremaps
             if (self.verbose and dead_neurons > 0) or dead_neurons_share > self.dead_neurons_share_threshold:
-                warnings.warn(
-                    'Layer #{} has {} dead neurons ({:.2%})!'
-                        .format(layer_index, dead_neurons, dead_neurons_share),
-                    RuntimeWarning
-                )
+                str_warning = 'Layer {} (#{}) has {} dead neurons ({:.2%})!'.format(layer_name, layer_index,
+                                                                                    dead_neurons, dead_neurons_share)
+                print(str_warning)
+                warnings.warn(str_warning, RuntimeWarning)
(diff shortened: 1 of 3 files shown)


infelane:
For testing I had 1 fail, however I'm quite certain it is independent of my additions.

`self = <backend_test.TestBackend object at 0x7f4f68019470>

    def test_moments(self):
        input_shape = (10, 10, 10, 10)
        x_0 = np.zeros(input_shape)
        x_1 = np.ones(input_shape)
        x_random = np.random.random(input_shape)
    
        th_axes = [0, 2, 3]
        tf_axes = [0, 1, 2]
    
        for ip in [x_0, x_1, x_random]:
            for axes in [th_axes, tf_axes]:
                for keep_dims in [True, False]:
                    ip_th = KTH.variable(ip)
                    th_mean, th_var = KCTH.moments(ip_th, axes, keep_dims=keep_dims)
    
                    ip_tf = KTF.variable(ip)
                    tf_mean, tf_var = KCTF.moments(ip_tf, axes, keep_dims=keep_dims)
    
                    th_mean_val = KTH.eval(th_mean)
                    tf_mean_val = KTF.eval(tf_mean)
                    th_var_val = KTH.eval(th_var)
                    tf_var_val = KTF.eval(tf_var)
    
                    assert_allclose(th_mean_val, tf_mean_val, rtol=1e-4)
>                   assert_allclose(th_var_val, tf_var_val, rtol=1e-4)
E                   AssertionError: 
E                   Not equal to tolerance rtol=0.0001, atol=0
E                   
E                   (mismatch 100.0%)
E                    x: array([[[[ 0.]],
E                   
E                           [[ 0.]],...
E                    y: array([[[[  6.938894e-18]],
E                   
E                           [[  6.938894e-18]],...

keras_contrib/callbacks/dead_relu_detector.py
```diff
@@ -25,7 +25,11 @@ def __init__(self, x_train, verbose=False):
 
     @staticmethod
     def is_relu_layer(layer):
-        return isinstance(layer, Dense) and layer.get_config()['activation'] == 'relu'
+        # Should work for all layers with relu activation. Tested for Dense and Conv2D
+        if 'activation' in layer.get_config():
+            return layer.get_config()['activation'] == 'relu'
+        else:
+            return False
```
arodiss: @infelane 
You can simplify this to
`return 'activation' in layer.get_config() and layer.get_config()['activation'] == 'relu'`

infelane: Indeed! How do I just recommit-repush-repull?

keras_contrib/callbacks/dead_relu_detector.py
```diff
@@ -44,17 +48,43 @@ def get_relu_activations(self):
...
+
+            # should work for both Conv and Flat
+            if K.backend() == 'tensorflow':
+                # features in last axis
+                axis_filter = -1
+            elif K.backend() == 'theano':
+                # features before the convolution axis, for weight_len the input and output have to be subtracted
+                axis_filter = -1 - (weight_len - 2)
+            else:
+                raise ValueError('Unknown backend: {}'.format(K.backend()))
```
arodiss: @infelane 
At the very least there is also NLTK backend (Microsoft's Natual Language Toolkit)
There are also some third-party back-ends, like plaidml: https://github.com/plaidml/plaidml
There may be more back-ends coming in the future

The bottom line is that ideally you should not care how the back-end is implemented in `keras`
Maybe there is a way to ask backend about axis order?

tests/keras_contrib/callbacks/dead_relu_detector_test.py
```diff
@@ -26,16 +37,112 @@ def do_test(weights, expected_warnings, verbose):
...
+def test_DeadDeadReluDetector_conv():
+    n_samples = 9
+
+    # (5, 5) kernel, 4 input featuremaps and 10 output featuremaps
+    if K.backend() == 'tensorflow':
+        input_shape = (n_samples, 5, 5, 4)
+    elif K.backend() == 'theano':
+        input_shape = (n_samples, 4, 5, 5)
+    else:
+        raise ValueError('Unknown backend: {}'.format(K.backend()))
```
arodiss: @infelane 
not good to rely on exact backend implementation

keras_contrib/callbacks/dead_relu_detector.py
infelane: K.image_data_format() should do the trick
Done

tests/keras_contrib/callbacks/dead_relu_detector_test.py
infelane: K.image_data_format() should do the trick
Done

keras_contrib/callbacks/dead_relu_detector.py
infelane: I think this was a necessary change. Even between TF and Theano you can still adjust the channel order. Making the previous check prone to fail if you were going to adjust the channel order manually.

keras_contrib/callbacks/dead_relu_detector.py
```diff
@@ -44,17 +45,41 @@ def get_relu_activations(self):
         layer_outputs = [func(list_inputs)[0] for func in funcs]
         for layer_index, layer_activations in enumerate(layer_outputs):
             if self.is_relu_layer(self.model.layers[layer_index]):
-                yield [layer_index, layer_activations]
+                layer_name = self.model.layers[layer_index].name
+                # layer_weight is a list [W] (+ [b])
+                layer_weight = self.model.layers[layer_index].get_weights()
+                # with kernel and bias, the weights are saved as a list [W, b]. If only weights, it is [W]
+                assert type(layer_weight) == list
```
ahundt: rather than an assert this should raise a ValueError

keras_contrib/callbacks/dead_relu_detector.py
```diff
@@ -44,17 +45,41 @@ def get_relu_activations(self):
...
+            dead_neurons_share = dead_neurons / total_featuremaps
             if (self.verbose and dead_neurons > 0) or dead_neurons_share > self.dead_neurons_share_threshold:
-                warnings.warn(
-                    'Layer #{} has {} dead neurons ({:.2%})!'
-                        .format(layer_index, dead_neurons, dead_neurons_share),
-                    RuntimeWarning
-                )
+                str_warning = 'Layer {} (#{}) has {} dead neurons ({:.2%})!'.format(layer_name, layer_index,
+                                                                                    dead_neurons, dead_neurons_share)
+                print(str_warning)
```
ahundt: should this print be here? make output matches how other keras calls handle it

infelane: I put the print there together with the warning since the warning stopped showing after some 2 epochs of dead neurons. Shall I just remove the print? It might have something to do with just my settings for warnings?
" "default" | print the first occurrence of matching warnings for each location where the warning is issued " https://docs.python.org/2/library/warnings.html#the-warnings-filter

arodiss: @ahundt you mean this should be logged to stderr? Makes sense IMHO

infelane: Okey, I fixed the problem at my side. The warning now shows everytime by setting:
>>>import warnings
>>>warnings.simplefilter("always")
or in terminal by : python python -W alwys foo.py

Now we only have to descide if we show a warning or just print it.

infelane: Or we give it as a setting:

>def foo(bool_warning = False):
>    if bool_warning:
>        warnings.warn(str_warning, RuntimeWarning)
>    else:
>        print(str_warning)

ahundt: Is warnings imported by keras? I'm just saying to double check how similar APIs in keras do things, maybe they just print most of the time then it is fine.

infelane: Yes, keras importants warnings in keras/callbacks.py. I think warnings is a standard library?

In my opinion a print will be saver. A warning is nicer. So I don't know if we should give the audience the choice between a print and a warning. I think this should be a decision from higher up (for example you :) )
Although if I look at other examples that use warnings. Here the warning isn't that crucial (unless ALL neurons are dead).

ahundt: Yes I looked at keras/callbacks.py, and based on that I think just a print would be fine.

keras_contrib/callbacks/dead_relu_detector.py
```diff
@@ -16,16 +15,22 @@ class DeadReluDetector(Callback):
             True means that even a single dead neuron triggers warning
             False means that only significant number of dead neurons (10% or more)
             triggers warning
+        bool_warning: output mode
+            True means a warning is raised
+            False means the warning message is printed.
     """
-    def __init__(self, x_train, verbose=False):
+
+    def __init__(self, x_train, verbose=False, bool_warning=False):
```
ahundt: can this become a string option with the variable name `on_detection` where accepted parameters are `'raise'` and `'print'`?

infelane: Do you think it is a good idea that the training should fail when a neuron is dead? If you wan't to prevent the neurons from dying I get that. But than we also have to be sure that this is checked before the network is saved.
At this point the printing works (with testing), I might update in a few minutes.

ahundt: Yes the other option is make this printing only, that's ok with me I'd be happy to merge if the param was gone entirely too.

infelane: I believe it might be a good idea to do something if a condition is met (e.g. a dead neuron). However this is more general. Do we agree to keep it to a print only for now?

ahundt: yep print is good. Thinking this through some more... Is there already a return value or some callback that lets the user detect the problem and act on it in their code? 

I don't think a print alone would let people detect a failure and, for instance, restart training in an automated fashion.

infelane:
If people need a return vallue/callback, the request will come naturally.

Now... My check failed multiple times. For a test that worked on my pc (both py2.7 Tensorflow and py3.5 Tensorflow) an never by the 'ravis CI build'. In the end I just commented that part of the test, but now it gives errors about an other package. I don't know what I can/should do about that

ahundt:
@infelane I took a look at the errors and it seems they aren't related to your changes. Does that seem correct to you?

infelane:
@ahundt The errors that raised were indeed outside my additions. I checked the code that raised the error and might have found a bug overthere. Do you think the dead_relu_detector can be merged? Maybe reopen the backend_test?

ahundt:
LGTM, merging, thanks for the contribution! 👍

arodiss:
Congrats for getting it merged

PR #180, opened 2017-11-17
Add NASNet models

titu1994:
Adds the model builder code for NASNet models - based on the paper [Learning Transferable Architectures for Scalable Image Recognition](https://arxiv.org/abs/1707.07012).

Contains pre-built models for NASNet-Large (6 @ 4032), NASNet-Mobile (4 @ 1056) and NASNet-CIFAR (6 @ 768), along with custom NASNet model builder.


titu1994:
@ahundt @farizrahman4u Is the test suite completely broken at this point? It shows that the tests didnt even begin at this point.

ahundt:
It should get to the tests now if you merge with master

ahundt:
Don't forget to include an example & simple test initializing the models.

titu1994:
@ahundt Seems like there are no application tests for some reason in contrib. Also, seems like CRF tests are failing.

ahundt:
yeah I noticed the CRF issue, but I don't know enough about it to make a good fix myself. A bunch of theano code is broken upstream in keras due to a new theano release (and probably here too).

titu1994:
@ahundt Hmm, should I merge this then ? I've tested it extensively locally, and it seems to train properly.

ahundt:
I'm looking through it, does it actually load the imagenet weights from an hdf5 file? I can't find them anywhere but it looks like they are the default

titu1994:
No it doesn't have the ImageNet weights. It simply warns that it was set, but that no weights are available so it continues. 

I should probably fix that.

ahundt:
Also any idea what people should see on cifar? Can a note be added? Sorry for so many litttle things

titu1994:
The paper simply mentions the final performance on all models and stresses on ImageNet. Not much on Cifar so I can't add anything for that sadly.

ahundt:
Sorry, wrt cifar I meant the pre-trained cifar weights expected performance.

titu1994:
I don't have the resources to train such models.

ahundt:
oh sorry I thought there was a pre-trained model. LGTM then!

PR #181, opened 2017-11-18
Adds swish activation function

valentindey:
Basically the same as #173 but allows for trainable beta value as described in the paper. Thus, I added it to advanced activations.


ahundt:
closing and reopening to re-run tests

ahundt:
@valentindey you might have to merge with master so we can ensure your changes don't introduce new test failures

ahundt:
LGTM, thanks!

PR #182, opened 2017-11-18
.travis.yml try fixing mkl errors

ahundt:


ahundt:
Just checked current failures, they're not due to mkl anymore. Merging this.

PR #185, opened 2017-12-02
.travis.yml attempt to fix travis error

ahundt:


ahundt:
Travis gets further along with coveralls removed and to my knowledge coveralls was never completely integrated anyway, so I'm merging this.

PR #186, opened 2017-12-04
Fix NASNet CIFAR and add Regularization

titu1994:
Fixes the NASNet model builder for CIFAR mode.

Also, adds the correct weight (l2) regularizers to the NASNet models, corrects BatchNormalization decay and epsilon values for CIFAR mode and corrects the example script.


examples/cifar10_nasnet.py
```diff
@@ -85,13 +83,21 @@
     # (std, mean, and principal components if ZCA whitening is applied).
     datagen.fit(X_train)
 
+    # wrap the ImageDataGenerator to yield two label batches [y, y] for each input batch X
```
ahundt: can you also add to the comment to explain why duplicate labels are needed? It seems unusual.

keras_contrib/applications/nasnet.py
```diff
@@ -252,19 +247,21 @@ def NASNet(input_shape=None,
         model = Model(inputs, x, name='NASNet')
 
     # load weights (when available)
-    warnings.warn('Weights of NASNet models have not been ported yet for Keras.')
+    if weights is not None:
+        warnings.warn('Weights of NASNet models have not been ported yet for Keras.')
```
ahundt: Should read:
`Weights of NASNet models have not yet been ported to Keras`

keras_contrib/applications/nasnet.py
```diff
@@ -618,36 +642,77 @@ def _reduction_A(ip, p, filters, id=None):
...
 
         with K.name_scope('block_5'):
             x5 = AveragePooling2D((3, 3), strides=(1, 1), padding='same', name='reduction_left5_%s' % id)(x1)
 
         x = concatenate([x2, x3, x5, x4], axis=channel_dim, name='reduction_concat_%s' % id)
         return x, ip
+
+
+def _add_auxilary_head(x, classes, weight_decay):
+    '''Adds an auxilary head for training the model
```
ahundt: Can you expand the comment to say why this is desirable or reference the page/section of the paper if it is from there?

titu1994:
@ahundt Made the changes as requested. The auxilary head is from the appendix of the paper and from the code base. It is used when training all of the models.

keras_contrib/applications/nasnet.py
titu1994: Added below.

examples/cifar10_nasnet.py
titu1994: Added below

PR #187, opened 2017-12-06
Add Weights for NASNet models

titu1994:
Contains changes to support weight files for NASNet Large and NASNet Mobile

Changes

- Small but important changes to the normal and reduction cells to make them correct
- Support for weight loading for 224 and 331 size inputs when built using the pre-built models

### DIFF
--- a/keras_contrib/applications/nasnet.py
+++ b/keras_contrib/applications/nasnet.py
@@ -186,7 +186,7 @@ def NASNet(input_shape=None,
             img_input = input_tensor
 
     assert penultimate_filters % 24 == 0, "`penultimate_filters` needs to be divisible " \
-                                          "by 6 * (2^N)."
+                                          "by 24."
 
     channel_dim = 1 if K.image_data_format() == 'channels_first' else -1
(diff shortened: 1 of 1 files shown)


titu1994:
@ahundt I believe this is now perfect. The weights are loading properly, and the changes ought to be the final major update. Sorry for all the small PRs, this is one hellish model to deal with.

ahundt:
Looks like it was very tricky! I don't think automated use of machine learning to design network models really considers the ease of human re-implementation. :-)

titu1994:
Haha true. This was probably the hardest model I've ever reimplenented in Keras. 

Thankfully, the weight loading mechanism is quite generic, so I can apply it to models in the future. It's just too labour intensive. 

For reference - https://github.com/titu1994/Keras-NASNet/tree/master/weight_translation

ahundt:
Woah cool, you should PR the transfer utilities that are generic here!

titu1994:
Well, it's not generic in that sense. You need to wrap the model with manual weights loading, provided by the manual loading of weights. 

Too much manual work really.

PR #188, opened 2017-12-07
backend_test.py remove scipy.sparse

ahundt:
remove scipy.sparse as it doesn't appear to be utilized in this file


ahundt:
🎉 🍾 Travis passes again! Merging

PR #192, opened 2017-12-15
TimeDistributed ResNet

voletiv:
Often, CNN+LSTM architectures require a TimeDistributed version of CNNs - one CNN architecture acting on several time-distributed inputs, and then fed into an LSTM.

Invoking the argument time_distributed=True shall create a TimeDistributed version of ResNet, with nb_time_steps versions of the ResNet, with the input_shape specified as (nb_time_steps, nb_rows, nb_cols, nb_channels).

### DIFF
--- a/keras_contrib/applications/resnet.py
+++ b/keras_contrib/applications/resnet.py
@@ -12,7 +12,9 @@ Reference material for extended functionality:
 - [Deep Residual Learning for Instrument Segmentation in Robotic Surgery](https://arxiv.org/abs/1703.08580)
   for segmentation support.
 
-Implementation Adapted from: github.com/raghakot/keras-resnet
+Implementation adapted from: github.com/raghakot/keras-resnet
+
+TimeDistributed adapted from: github.com/voletiv/keras-resnet
 """
 from __future__ import division
 
@@ -26,10 +28,10 @@ from keras.layers import Flatten
 from keras.layers import Conv2D
 from keras.layers import MaxPooling2D
 from keras.layers import AveragePooling2D
-from keras.layers.pooling import GlobalAveragePooling2D
 from keras.layers import GlobalMaxPooling2D
 from keras.layers import GlobalAveragePooling2D
 from keras.layers import Dropout
+from keras.layers import TimeDistributed
 from keras.layers.merge import add
 from keras.layers.normalization import BatchNormalization
 from keras.regularizers import l2
@@ -37,10 +39,15 @@ from keras import backend as K
 from keras.applications.imagenet_utils import _obtain_input_shape
 
 
-def _bn_relu(x, bn_name=None, relu_name=None):
+def _bn_relu(x, bn_name=None, relu_name=None, time_distributed=False, verbose=False):
     """Helper to build a BN -> relu block
     """
-    norm = BatchNormalization(axis=CHANNEL_AXIS, name=bn_name)(x)
+    if verbose:
+        print("    _bn_relu")
+    if time_distributed:
+        norm = TimeDistributed(BatchNormalization(axis=CHANNEL_AXIS))(x)
+    else:
+        norm = BatchNormalization(axis=CHANNEL_AXIS, name=bn_name)(x)
     return Activation("relu", name=relu_name)(norm)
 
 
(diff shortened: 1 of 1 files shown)


titu1994:
I do not believe this is the correct approach. For CNN + LSTM models, the correct approach is as below : 
```python
from keras.layers import TimeDistributed
from keras_contrib.applications.resnet import ResNet

input_layer = Input(shape=(Timedim, rows, cols, channels))
time_distributed_output = TimeDistributed(ResNet(...))(input_layer)

model = Model(input_layer, time_distributed_output)
```

There is no reason to use TimeDistributed over each block.

However, since @ahundt is the author of the ResNet models in Keras, I will defer the final decision to him.

voletiv:
@titu1994  I had tried that approach but it hadn't worked. That's why I had to wrap TimeDistributed individually over all blocks except the skip connections. 

I agree, let's defer the decision to @ahundt, he was the one who had suggested this PR to me.

ahundt:
I'm very interested in this feature, thanks for the PR!

@voletiv do you have test code that can reproduce the issue with what @titu1994 suggested? 
@titu1994 do you have any suggestions that might resolve the problem?

A side note is that I've also got a few PRs open for this code, and I think @roatienza identified a bug mentioned in https://github.com/keras-team/keras/pull/8231#issuecomment-338896620. @voletiv have you encountered that problem? 
@roatienza do you perhaps have additional details?

keras_contrib/applications/resnet.py
```diff
@@ -113,27 +147,45 @@ def _shortcut(input_feature, residual, conv_name_base=None, bn_name_base=None):
...
-        shortcut = Conv2D(filters=residual_shape[CHANNEL_AXIS],
-                          kernel_size=(1, 1),
-                          strides=(stride_width, stride_height),
-                          padding="valid",
-                          kernel_initializer="he_normal",
-                          kernel_regularizer=l2(0.0001),
-                          name=conv_name_base)(input_feature)
+        if time_distributed:
+            shortcut = TimeDistributed(Conv2D(filters=residual_shape[CHANNEL_AXIS],
+                                              kernel_size=(1, 1),
```
ahundt: can there just be one call to Conv2D here and then an if statement that adds TimeDistributed below if enabled?

keras_contrib/applications/resnet.py
```diff
@@ -113,27 +147,45 @@ def _shortcut(input_feature, residual, conv_name_base=None, bn_name_base=None):
     shortcut = input_feature
     # 1 X 1 conv if shape is different. Else identity.
     if stride_width > 1 or stride_height > 1 or not equal_channels:
-        print('reshaping via a convolution...')
+        if verbose:
+            print("    SHORTCUT not right: input -", shortcut.shape, "residual -", residual.shape)
```
ahundt: I like this expanded comment. Could it include a suggestion on what a user can do to fix the problem?

keras_contrib/applications/resnet.py
```diff
@@ -173,99 +227,157 @@ def _block_name_base(stage, block):
...
             # don't repeat bn->relu since we just did bn->relu->maxpool
-            x = Conv2D(filters=filters, kernel_size=(3, 3),
-                       strides=transition_strides,
-                       dilation_rate=dilation_rate,
-                       padding="same",
-                       kernel_initializer="he_normal",
-                       kernel_regularizer=l2(1e-4),
-                       name=conv_name_base + '2a')(input_features)
+            if time_distributed:
+                x = TimeDistributed(Conv2D(filters=filters, kernel_size=(3, 3),
```
ahundt: same as above, change to a single convolution and add the TimeDistributed component below. Will the two calls to TimeDistributed interfere with each other?

titu1994:
@ahundt I don't really work with CNN LSTM models, so I thought that version could work (it was referenced by @fchollet in one of the presentations of Tensorflow, though I can't recall the title of the speech at the moment. 

It would seem to work for Inception v3, so I assumed it would apply to ResNets as well. It would be interesting to know if this wasn't the case for some reason.

keras_contrib/applications/resnet.py
voletiv: This is not a user-related problem, there is nothing to be fixed.

If the input_feature and residual have unequal channels, then the skip connection cannot be made. In this case, the input_feature is transformed to have the same number of channels as the residual and they are added to make a skip connection.

keras_contrib/applications/resnet.py
voletiv: Great suggestion, will implement and squash commits.

keras_contrib/applications/resnet.py
voletiv: Great suggestion, will implement and squash commits.

voletiv:
I used the following code (at the root folder of keras-contrib) to show TImeDistributed(ResNet) does not work properly. Although the error is related to BatchNorm, probing into it further seemed more difficult than making this PR!

```
import numpy as np

from keras_contrib.applications.resnet import *
from keras.layers import TimeDistributed

time_steps = 8
input_shape = (128, 128, 3)
channels = 10

# Making the naive TimeDistributed version of ResNet
input_tensor = Input(shape=(time_steps, *input_shape))
resnet_model = ResNet18(input_shape, channels)
output_tensor = TimeDistributed(resnet_model)(input_tensor)
naive_time_distributed_resnet_model = Model(inputs=[input_tensor], outputs=[output_tensor])

# To test
a = np.random.random((2, time_steps, *input_shape))

# Testing
b = naive_time_distributed_resnet_model.predict(a)

# We get the following error:
# InvalidArgumentError (see above for traceback): You must feed a value for placeholder tensor 'batch_normalization_1/keras_learning_phase' with dtype bool

# My time_distributed version
my_time_distributed_resnet_model = ResNet18((time_steps, *input_shape), channels, time_distributed=True)

# Testing
b = my_time_distributed_resnet_model.predict(a)

b.shape
# (2, 8, 10)
```

Let me know what you think.

roatienza:
When I was implementing the keras cifar10_resnet.py, there were some architectural differences from the keras-contrib resnet and the paper that I discovered. This is why keras-contrib version could not reach accuracy >90 for the shallow resnet during the time I tested it.

ahundt:
@voletiv I looked into it and to the best of my knowledge I believe what you've done is the best way to go about this task. However, there were a number of bugs in the code, mostly due to existing bugs. 

Can you review #167 and test it out to see if it works for you? I've incorporated your commits there.

ahundt:
@voletiv Can you also try the code you posted above with #167 to confirm that approach still doesn't work? If that's the case I think we should produce a bug report issue upstream on Keras.

voletiv:
I've reviewed #167, my changes have been faithfully reproduced, thanks! I have made a few comments about other minor bugs.

I tried the code I posted above with the updated code in #167, naive way still didn't work, mine did.

In test_resnet.py, maybe include a prediction as well, like I did in the code above? Because naive_time_distributed_resnet_model compiles without hiccups, so mere compilation might not be a good metric for testing.

voletiv:
@roatienza Is it possible the increase in accuracy is simply because of mean subtraction? (which wasn't done in the original file w.r.t. your PR)

ahundt:
closing in favor of #167 which contains these changes

PR #195, opened 2017-12-18
Segmentation Data Generator

ahundt:
This generates data for semantic segmentation tasks, which specifically need second images for labels, for which the pixel values cannot change from a fixed set, i.e. no interpolation because it changes the ground truth label values.

source: https://github.com/aurora95/Keras-FCN


wassname:
With the latest version of keras and keras-contrib I get "ValueError: too many values to unpack (expected 3) "

```
    184         with self.lock:
    185             index_array, current_index, current_batch_size = next(
--> 186                 self.index_generator)
    187 
    188         # The transformation of images is not under thread lock so it can be

ValueError: too many values to unpack (expected 3)
```

I think this is because flow index is returning a single value but it's expecting three [here](https://github.com/farizrahman4u/keras-contrib/blob/segmentation-data-generator/keras_contrib/preprocessing/image_segmentation.py#L185)

ahundt:
hmm, looks like it is due to changes in https://github.com/keras-team/keras/blob/master/keras/preprocessing/image.py#L767

wassname:
Ah yeah looks like it changed from 2.0.8 to 2.0.9+, so this isn't a problem in 2.0.8

ahundt:
yeah... I started trying to fix it where it is fully integrated already but ran into another issue:
https://github.com/aurora95/Keras-FCN/pull/56

ahundt:
@wassname OK I think it should work now with 2.1.x

wassname:
For me I got `SegDirectoryIterator has no attribute next`. So you might want to add a next method like in https://github.com/keras-team/keras/blob/master/keras/preprocessing/image.py#L1161 DirectoryIterator and NumpyArrayIterator.

ahundt:
closing in favor of https://github.com/keras-team/keras-contrib/pull/202 because this PR was accidentally made as a branch directly in keras-contrib, rather than github.com/ahundt/keras-contrib.

PR #198, opened 2017-12-20
Add WarmRestart callback

nzw0301:
Warm Restart is a technique to update learning rate. I think that it is useful to add this as the new callback function for Keras users.

## TODO

- [x] Add test based on `test_LearningRateScheduler` in [callback test code](https://github.com/keras-team/keras/blob/master/tests/keras/test_callbacks.py).
- [x] Show learning process plot that is similar to original paper.
- [x] Improve docstring

---

Reference: Ilya Loshchilov, Frank Hutter. SGDR:   Stochastic Gradient Descent with Warm Restarts. In ICLR, 2017. [paper](https://arxiv.org/pdf/1608.03983.pdf).

### DIFF
--- a/tests/keras_contrib/callbacks/warm_restart_test.py
+++ b/tests/keras_contrib/callbacks/warm_restart_test.py
@@ -1,0 +1,42 @@
+from keras.models import Sequential
+from keras.layers.core import Dense
+from keras.utils.test_utils import get_test_data
+from keras import backend as K
+from keras.utils import np_utils
+from keras_contrib import callbacks
+
+import numpy as np
+import pytest
+
+input_dim = 2
+num_hidden = 4
+num_classes = 2
+batch_size = 5
+train_samples = 20
+test_samples = 20
+
+
+def LearningRateWarmRestarter():
+    np.random.seed(1337)
+    (X_train, y_train), (X_test, y_test) = get_test_data(num_train=train_samples,
+                                                         num_test=test_samples,
+                                                         input_shape=(input_dim,),
+                                                         classification=True,
+                                                         num_classes=num_classes)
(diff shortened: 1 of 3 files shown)


nzw0301:
Oh... I will fix it also.
https://travis-ci.org/farizrahman4u/keras-contrib/jobs/318981030

ahundt:
looks nice! could a unit test be added and the docstring expanded with when/why to use it and what losses it is compatible with?

Ideally, an example would be great too if you have a chance but not an absolute requirement.

nzw0301:
@ahundt Thank you for your comment! I will update this docstring more detail. 
Ok, I will add code snippet as this callback's example.

tests/keras_contrib/callbacks/warm_restart_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+input_dim = 2
+num_hidden = 4
+num_classes = 2
+batch_size = 5
+train_samples = 20
+test_samples = 20
+
+
+def LearningRateWarmRestarter():
```
ahundt: could this be named something other than the class name in warm_restart.py?

nzw0301: done. Thanks!
`LearningRateWarmRestarter` -> `test_LearningRateWarmRestarter`

nzw0301:
I tested this new feature for CNN on CIFAR-10.
https://gist.github.com/nzw0301/66d88fc81687cab6856ad2f1a947383e

Unfortunately, its validation error is worse than no warm restart version.
So, I close this PR once. If I identify the reason not to work well, I will re-open this PR.
Sorry.

index

ahundt:
Ok, thanks for the accurate assessment :+1:

PR #208, opened 2018-01-05
add logsumexp in tensorflow_backend.py

JeffpanUK:
In crf.py there is a calling for backend.logsumexp. When using tensorflow as backend, an error will occur as there is no such a function in keras_contrib.backend tensorflow.py.

### DIFF
--- a/keras_contrib/backend/cntk_backend.py
+++ b/keras_contrib/backend/cntk_backend.py
@@ -1,4 +1,5 @@
 from keras.backend import cntk_backend as KCN
+from keras.backend.cntk_backend import logsumexp
 import cntk as C
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


ahundt:
Wouldn't a simpler solution be to [import it from keras' tensorflow_backend.py](https://github.com/keras-team/keras/blob/04c11f8c13b1182b43eaed36dbfdedb23efd38b9/keras/backend/tensorflow_backend.py#L1485)?

`from keras.backend import logsumexp`

JeffpanUK:
just updated the pull request. removing the logsumexp function and import it from keras.tensorflow_backend directly.

ahundt:
Great, thanks! Since this is clearly already broken in master (and travis is broken too), I'll merge this if you can confirm you were able to run a test locally.

JeffpanUK:
Hi @ahundt , I've tested it locally and I can confirm it works. I also removed some space lines to follow the pep8.

PradyumnaGupta:
The current tensorflow.keras.backend does not have logsumexp and hence the code is not working.

PR #225, opened 2018-02-09
Depth Typo for densenet 264

ps48:


ahundt:
@titu1994 can you confirm this?

titu1994:
Sorry for the delay. The fix is correct. It doesn't affect functionality, since the depth argument is overridden by the other parameters when custom nb_layers_per_block is defined.

ahundt:
Thanks!

Issue #222, opened 2018-02-06
Creating a resnet-152 -> exception when naming block


oliviersoares:
If you create a big resnet (resnet-152 for e.g.), _block_name_base in file "keras_contrib/applications/resnet.py" will throw an exception:
```
conv_name_base = 'res' + str(stage) + block + '_branch'
TypeError: cannot concatenate 'str' and 'int' objects
```

If block < 27, its type will be converted to 'str', and if >= 27, its type will stay 'int'.

To fix it, replace:
```
    if block < 27:
        block = '%c' % (block + 97)  # 97 is the ascii number for lowercase 'a'
```

with something like this:
```
    if block < 27:
        block = '%c' % (block + 97)  # 97 is the ascii number for lowercase 'a'
    else:
        block = '%d' % (block - 27)
```

oliviersoares:
Btw, it should be "26", not "27". '%c' (26 + 97) is character '{'.
You want to stop at 25 with character 'z'.

Issue #221, opened 2018-02-06
resnets with repetition > 2 are failing (i.e. all resnets except ResNet-18)


oliviersoares:
If you run the following code:

from keras_contrib.applications.resnet import ResNet18, ResNet101
cnn1 = ResNet18((224, 224, 3), 1000)
cnn2 = ResNet101((224, 224, 3), 1000)

The first resnet is created successfully, however the second one is failing with this exception:

  File "~/venv/local/lib/python2.7/site-packages/keras_contrib/applications/resnet.py", line 454, in ResNet101
    return ResNet(input_shape, classes, bottleneck, repetitions=[3, 4, 23, 3])
  File "~/venv/local/lib/python2.7/site-packages/keras_contrib/applications/resnet.py", line 403, in ResNet
    residual_unit=residual_unit)(block)
  File "~/venv/local/lib/python2.7/site-packages/keras_contrib/applications/resnet.py", line 159, in f
    dilation_rate=dilation_rates[i],
IndexError: tuple index out of range

Looking quickly at the code, any resnet using a repetition > 2 is failing as dilation_rates has a size of 2 by default.

You can easily fix the code by initializing in the _residual_block function the parameter "dilation_rates" to None (instead of (1, 1)) and adding at the beginning of the function:
```
    if dilation_rates is None:
        dilation_rates = [1] * blocks
```

SSchott:
Found the same issue. Could you generate a pull request for this?

adambielski:
still an issue...

SSchott:
Sadly yes, but either add it to the code by yourself or get the one I added. Is one line anyway, so shouldn't be such a big issue.

PR #226, opened 2018-02-09
Updating ror for require_flatten error and changes to resnet.py as per recent issues

anand-sonawane:


anand-sonawane:
fixes #222 and #221

keras_contrib/applications/resnet.py
```diff
@@ -132,7 +132,7 @@ def _shortcut(input_feature, residual, conv_name_base=None, bn_name_base=None):
 
 def _residual_block(block_function, filters, blocks, stage,
                     transition_strides=None, transition_dilation_rates=None,
-                    dilation_rates=(1, 1), is_first_layer=False, dropout=None,
+                    dilation_rates=None, is_first_layer=False, dropout=None,
```
ahundt: This dilation rate is a 2D dimensional multiplier. It is not a list per block. I don't think this change is correct

anand-sonawane: Yep, its a 2D dimensional multiplier, I was trying it as a list as well and it seems to work which confused me. I have made the change, the change was required because initializing a Deep Resnet was throwing a error with the current code.

anand-sonawane:
@ahundt can you please look into this now? Because the error still exists.

anand-sonawane:
@ahundt Please make this change as it still persists or close the issue.

ahundt:
@anand-sonawane can you fix the conflict?

gabrieldemarmiesse:
I'll be closing this PR as it has conflicts and they were not resolved for quite some time. It should probably be easier to open a new one at this point.

PR #227, opened 2018-02-09
Fixed CRF incompatibilities with keras 2.1.3

DanAnastasyev:
The problem was introduced by [this commit](https://github.com/keras-team/keras/commit/958239c621a6be037c5f8b30be9270310735f725)


ahundt:
Awesome, LGTM, thanks!

PR #232, opened 2018-02-21
fix the gradient penalty loss in the improved WGAN example

DavidS3141:
# Issue
Fixed the computation of the gradient penalty loss in the WGAN example according to the [related paper](https://arxiv.org/abs/1704.00028), citing the following loss:
gp =  lambda * average( ( 1 - ||gradient_i|| )^2 )

The old implementation did compute the following:
gp = lambda * ( 1 - sqrt(sum(||gradient_i||^2)) )^2

# Differences at execution
Old version leads to a comparable loss as the new (according to the paper correct) version and to train gradients to the value ||gradient_i|| = 1 / sqrt(N) where i=1...N with N the batch size. So the resulting critic does not compute the earth-movers distance(EMD) but EMD / sqrt(N). These are the most significant changes and the reason why the old version is still working and converging, but depends on the batch size which is not very nice to have.

# Testing
I tested this with the backend TensorFlow and it works as expected.


examples/improved_wgan.py
```diff
@@ -71,10 +71,23 @@ def gradient_penalty_loss(y_true, y_pred, averaged_samples, gradient_penalty_wei
...
-    gradients = K.gradients(K.sum(y_pred), averaged_samples)
-    gradient_l2_norm = K.sqrt(K.sum(K.square(gradients)))
+    # first get the gradients:
+    #   assuming: - that y_pred has dimensions (batch_size, 1)
+    #             - averaged_samples has dimensions (batch_size, nbr_features)
+    # gradients afterwards has dimension (batch_size, nbr_features), basically
+    # a list of nbr_features-dimensional gradient vectors
+    gradients = K.gradients(y_pred, averaged_samples)
+    # compute the euclidean norm by squaring ...
+    gradients_sqr = K.square(gradients)[0]
```
lgeiger: I think we should select the first element of the gradient list before taking the square root:
```python
gradients = K.gradients(y_pred, averaged_samples)[0]
gradients_sqr = K.square(gradients)
```
Maybe a *tiny* performance improvement 🚀

DavidS3141: Oh that's actually the square function adding this dimension for some reason. Pretty weird but your suggestion would actually change the behaviour.

lgeiger: > Oh that's actually the square function adding this dimension for some reason.

`K.square` should compute the square element-wise so I'm pretty sure it doesn't change the shape: https://keras.io/backend/#square
Or am I missing something?

DavidS3141: Ah you were right ;) I encountered this problem already in my own implementation and thought it was because of some strange implementation of the square function, but didn't gave it too much thought. Didn't noticed that actually the gradients function returns a list of tensors, this list being the additional dimension.

examples/improved_wgan.py
```diff
@@ -71,10 +71,23 @@ def gradient_penalty_loss(y_true, y_pred, averaged_samples, gradient_penalty_wei
...
     y_true and y_pred to loss functions. To get around this, we make a partial() of the function with the
     averaged_samples argument, and use that for model training."""
-    gradients = K.gradients(K.sum(y_pred), averaged_samples)
-    gradient_l2_norm = K.sqrt(K.sum(K.square(gradients)))
+    # first get the gradients:
+    #   assuming: - that y_pred has dimensions (batch_size, 1)
+    #             - averaged_samples has dimensions (batch_size, nbr_features)
+    # gradients afterwards has dimension (batch_size, nbr_features), basically
+    # a list of nbr_features-dimensional gradient vectors
+    gradients = K.gradients(y_pred, averaged_samples)[0]
```
ahundt: can you explain why only index 0 is accessed?

DavidS3141: gradients(y, [x1, x2, ...]) always returns a list, each entry is the gradient of y with respect to xi.
As we only pass one tensor for x (averaged_samples) and gradients can handle non-wrapped single elements the output is still in list form, but only containing this one elment. As we don't want the list but the actual element, we need the indexing.

PR #237, opened 2018-03-03
nasnet.py add sigmoid activation option

ahundt:
@titu1994 could you review this change?

Also, I found a couple bugs in the NASNet weights:

1. It seems the actual NASNet weights you posted have a misspelling of auxiliary as "auxilary":
```
NASNET_LARGE_WEIGHT_PATH_WITH_auxiliary_NO_TOP = "https://github.com/titu1994/Keras-NASNet/releases/download/v1.1/NASNet-auxilary-large-no-top.h5"
```

2. Additionally, when you go to load the above weights, it seems the auxiliary network's "top" weights are still present, so if the number of classes has changed loading of the weights fails.
3. The auxiliary network should probably respect choices for the top, pooling, etc

### DIFF
--- a/keras_contrib/applications/nasnet.py
+++ b/keras_contrib/applications/nasnet.py
@@ -53,7 +53,7 @@ NASNET_MOBILE_WEIGHT_PATH_WITH_AUXULARY_NO_TOP = "https://github.com/titu1994/Ke
 NASNET_LARGE_WEIGHT_PATH = "https://github.com/titu1994/Keras-NASNet/releases/download/v1.1/NASNet-large.h5"
 NASNET_LARGE_WEIGHT_PATH_NO_TOP = "https://github.com/titu1994/Keras-NASNet/releases/download/v1.1/NASNet-large-no-top.h5"
 NASNET_LARGE_WEIGHT_PATH_WITH_auxiliary = "https://github.com/titu1994/Keras-NASNet/releases/download/v1.1/NASNet-auxiliary-large.h5"
-NASNET_LARGE_WEIGHT_PATH_WITH_auxiliary_NO_TOP = "https://github.com/titu1994/Keras-NASNet/releases/download/v1.1/NASNet-auxiliary-large-no-top.h5"
+NASNET_LARGE_WEIGHT_PATH_WITH_auxiliary_NO_TOP = "https://github.com/titu1994/Keras-NASNet/releases/download/v1.1/NASNet-auxilary-large-no-top.h5"
 
 
 def NASNet(input_shape=None,
@@ -222,13 +222,15 @@ def NASNet(input_shape=None,
     auxiliary_x = None
     if not skip_reduction:  # imagenet / mobile mode
         if use_auxiliary_branch:
-            auxiliary_x = _add_auxiliary_head(x, classes, weight_decay, activation=activation)
+            auxiliary_x = _add_auxiliary_head(x, classes, weight_decay, include_top=include_top,
+                                              pooling=pooling, activation=activation)
 
     x, p0 = _reduction_A(x, p, filters * filters_multiplier ** 2, weight_decay, id='reduce_%d' % (2 * nb_blocks))
 
     if skip_reduction:  # CIFAR mode
         if use_auxiliary_branch:
-            auxiliary_x = _add_auxiliary_head(x, classes, weight_decay, activation=activation)
+            auxiliary_x = _add_auxiliary_head(x, classes, weight_decay, include_top=include_top,
+                                              pooling=pooling, activation=activation)
 
     p = p0 if not skip_reduction else p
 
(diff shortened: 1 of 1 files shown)


titu1994:
That's bad. The weight loading scheme is finicky. I'll see what I can do about it. 

As to this change, it's perfect.

titu1994:
About the auxiliary networks adapting to top and pooling, I don't agree that they should be changed. 

The auxiliary branch is a regularizer, it has no purpose during evaluation and therefore should not have the specific modifier for general classifiers.

titu1994:
Hmm, fair point. Could you handle this if you don't mind ? I'm swamped with coursework at the moment.

ahundt:
I think my second commit should fix the issues while default behavior remains the same. However, perhaps it is worthwhile to add a renamed weights to your releases so we can fix that in one additional commit.

titu1994:
Right. I'll look into the weights over the weekend. Or probably tonight after I return. I'll inform here once I've worked out the issue.

ahundt:
Nope there's some kind of bug:
```
nasnet x: Tensor("single_hypertree/activation_263/Relu:0", shape=(?, 11, 11, 4032), dtype=float32) aux x: Tensor("single_hypertree/auxiliary_branch/activation_179/Relu:0", shape=(?, 1, 1, 768), dtype=float32)
```

Is this difference in dimension between the aux network and the main network expected? I'm not sure how to address the problem yet but taking a look.

ahundt:
Oh wait no after a second look that's expected behavior because I have pooling disabled. Hopefully this should be good to go!

titu1994:
> Additionally, when you go to load the above weights, it seems the auxiliary network's "top" weights are still present, so if the number of classes has changed loading of the weights fails.

Is this referring to the weights of the auxiliary classifier? This is a tad more involved to resolve. First the model needs to be patched to respect the pooling and include_top parameters for the auxiliary classifier.

For now, I have updated the weight names to conform to the urls here, so no need to update the paths in the file.

As to this case, I will be patching the above fix locally in my repository and then pushing the updated weight files. At your end, you would need to do the update the `_add_auxiliary_head(...)` function with the below:

```python
def _add_auxiliary_head(x, classes, weight_decay, pooling, include_top):
    '''Adds an auxiliary head for training the model

    From section A.7 "Training of ImageNet models" of the paper, all NASNet models are
    trained using an auxiliary classifier around 2/3 of the depth of the network, with
    a loss weight of 0.4

    # Arguments
        x: input tensor
        classes: number of output classes
        weight_decay: l2 regularization weight

    # Returns
        a keras Tensor
    '''
    img_height = 1 if K.image_data_format() == 'channels_last' else 2
    img_width = 2 if K.image_data_format() == 'channels_last' else 3
    channel_axis = 1 if K.image_data_format() == 'channels_first' else -1

    with K.name_scope('auxiliary_branch'):
        auxiliary_x = Activation('relu')(x)
        auxiliary_x = AveragePooling2D((5, 5), strides=(3, 3), padding='valid', name='aux_pool')(auxiliary_x)
        auxiliary_x = Conv2D(128, (1, 1), padding='same', use_bias=False, name='aux_conv_projection',
                            kernel_initializer='he_normal', kernel_regularizer=l2(weight_decay))(auxiliary_x)
        auxiliary_x = BatchNormalization(axis=channel_axis, momentum=_BN_DECAY, epsilon=_BN_EPSILON,
                                        name='aux_bn_projection')(auxiliary_x)
        auxiliary_x = Activation('relu')(auxiliary_x)

        auxiliary_x = Conv2D(768, (auxiliary_x._keras_shape[img_height], auxiliary_x._keras_shape[img_width]),
                            padding='valid', use_bias=False, kernel_initializer='he_normal',
                            kernel_regularizer=l2(weight_decay), name='aux_conv_reduction')(auxiliary_x)
        auxiliary_x = BatchNormalization(axis=channel_axis, momentum=_BN_DECAY, epsilon=_BN_EPSILON,
                                        name='aux_bn_reduction')(auxiliary_x)
        auxiliary_x = Activation('relu')(auxiliary_x)

        if include_top:
            auxiliary_x = GlobalAveragePooling2D()(auxiliary_x)
            auxiliary_x = Dense(classes, activation='softmax', kernel_regularizer=l2(weight_decay),
                                name='aux_predictions')(auxiliary_x)
        else:
            if pooling == 'avg':
                auxiliary_x = GlobalAveragePooling2D()(auxiliary_x)
            elif pooling == 'max':
                auxiliary_x = GlobalMaxPooling2D()(auxiliary_x)
                
    return auxiliary_x

```

and finally pass `pooling, include_top` to the `_add_auxiliary_head` function calls in the NASNet(...) function

titu1994:
Update: New weights for NASNet Large and Mobile with no top and auxiliary branch have been pushed now.

PR #243, opened 2018-03-28
Addition of GroupNormalization

titu1994:
Adds Group Normalization + associated tests from the paper [Group Normalization](https://arxiv.org/abs/1803.08494)

By extention, setting the parameter "groups" to 1 should make it equivalent to Instance Normalization and settng "groups" to the number of input channels will apply layer normalization.

### DIFF
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -142,6 +142,7 @@ class InstanceNormalization(Layer):
         base_config = super(InstanceNormalization, self).get_config()
         return dict(list(base_config.items()) + list(config.items()))
 
+
 get_custom_objects().update({'InstanceNormalization': InstanceNormalization})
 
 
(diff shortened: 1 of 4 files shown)


titu1994:
CNTK tests fail due to library issues, however the Keras code for mean and standard calculation remains the same for all backends, so I feel it can be skipped in the short term.

ahundt:
This looks nice. I bet it is something that is comparatively likely to be accepted into Keras upstream. Should we check for interest there first?

titu1994:
@ahundt Sure, I'll add a request there. For the time being however, I think I'll have to switch to manually computing the mean and variance if I push to Keras. It doesnt have the equivalent K.moments() op, which we added to Contrib quite some time ago. 

There are performance repercussions in this implementation. While it follows the code from the paper quite closely (with some modifications for how Keras handles different data formats etc), K.batch_normalization() and the internal code of batchnorm checks for and uses fused batchnorm ops when available. 

Those won't be available for either contrib or Keras proper until the appropriate ops are added to TF itself (and all this is under the hope that GroupNormalization gains enough favor in the community to warrant an additional fused op in TF). 

Edit: I forgot to say, I think we can keep this version here, since it uses the contrib `moments` op anyway.

tests/keras_contrib/backend/backend_test.py
```diff
@@ -151,14 +153,21 @@ def test_moments(self):
...
                     tf_mean_val = KTF.eval(tf_mean)
+                    # cntk_mean_val = KCTK.eval(cntk_mean)
                     th_var_val = KTH.eval(th_var)
                     tf_var_val = KTF.eval(tf_var)
+                    # cntk_var_val = KCTK.eval(cntk_var)
 
                     # absolute tolerance needed when working with zeros
                     assert_allclose(th_mean_val, tf_mean_val, rtol=1e-4, atol=1e-10)
                     assert_allclose(th_var_val, tf_var_val, rtol=1e-4, atol=1e-10)
+                    # assert_allclose(th_mean_val, cntk_mean_val, rtol=1e-4, atol=1e-10)
```
ahundt: should probably correct the commented lines in this file

titu1994:
Can't builds won't pass, either here or in Keras proper. I'm investigating the reason currently but have no plausible source of error yet.

0x00b1:
Thanks, @titu1994. I started merging this into 0.2.0 of [Keras-ResNet](https://github.com/broadinstitute/keras-resnet):

https://github.com/broadinstitute/keras-resnet/pull/40

Would you mind if I added you to the AUTHORS.rst?

titu1994:
Sure, I don't mind, but know that grouped convolutions as I have implemented here do not work on CNTK backend. TF and Theano show no problems, and I've been using this frequently for the past few weeks.

I would say, unless you are fine tuning a task, or have small batch sizes, this is inferior to batch norm + 128 batch size. That's just my observation on very select tests though so I won't comment on overall benefit.

0x00b1:
@titu1994 I’ll work on fixing the CNTK issues. I’ve actually seen worse results with typical batch sizes, so it won’t replace `BatchNormalization` as the default normalization layer. However, the Keras-ResNet is used by a few packages for image segmentation and object detection where batch sizes are usually one or two and group normalization performs **dramatic** performance benefits.

titu1994:
Yeah I've noticed similar results. It's surprisingly powerful for small batch sizes, even beating instance norm with just groups of 2. It's weird that it would have such a significant difference. 

I'm any case, I've been using it as a drop in replacement for instance norm in a few style transfer tasks, and it performs on par with actual instance norm at 16 batch size so it's useful enough for me.

titu1994:
Btw, if you do get a fix for the CNTK issues, please send a fix here as well.

0x00b1:
@titu1994 Of course. 👍 Ideally, it won’t stay in Keras-ResNet for too long.

keras_contrib/layers/normalization.py
```diff
@@ -373,4 +374,186 @@ def get_config(self):
...
+
+class GroupNormalization(Layer):
+    """Group normalization layer
+
+    Group Normalization divides the channels into groups and computes within each group
+    the mean and variance for normalization. GN's computation is independent of batch sizes,
+    and its accuracy is stable in a wide range of batch sizes
+
+    # Arguments
+        groups: Integer, the number of groups for Group Normalization.
```
ahundt: Can this include advice on how to choose the integer?

tests/keras_contrib/backend/backend_test.py
titu1994: The issue is that CNTK tests won't pass for some reason. It is using Contrib's `K.moments()` call, and technically, it should do just fine on CNTK, but it fails. And now that Contribs tests are broken, we cant check properly either.

keras_contrib/layers/normalization.py
titu1994: Hmm, I can mention that setting groups to 1 makes this identical to Layer Normalization, and setting it to the number of input channels makes this Instance Normalization?

titu1994: Potential fix, use the private `_moments` method directly from the CNTK backend. It will be more fragile to internal Keras changes, but I think moments op is too general to have any major changes.

titu1994:
@ahundt I may have a potential solution to the CNTK issue, which is to use the private backend function `_moments` from the CNTK backend directly and wrap it in `KC.moments()`.

Also, added the comments about how altering `groups` will affect the behavior of the normalization.

titu1994:
@ahundt I dunno if its just my setup or it's an actual issue, but Keras 2.2.0 combined with TF 1.8 always segfaults when I use K.clear_session() and follow up with building another model and then calling fit.

Perhaps that is another source of test errors. I feel we should create a PR that mimics Keras and get these tests passing at least.

titu1994:
@ahundt This is ready for merging as well, and passes CNTK tests now.

Are there any other corrections I should make, or can we merge this?

PR #247, opened 2018-04-12
Improved TimeseriesGenerator

TanguyUrvoy:
**TimeseriesGenerator**

- length is now the length of the produced samples (which means that we now sample on a window of size length*sample_rate). This fixes a bug: previous  TimeseriesGenerator did not work when sample_step was not a multiple of length.
- if not specified, output dtype is now infered from data and target (an integer array give integer arrays). This is helpful for integer sequences (like in RNN language models).
- data and targets are cast to numpy arrays and len(data) is no more used to set max_index (as a consequence, it should also works with multivariate time-series).
- shuffle is now a real shuffle (i.e. a sampling without replacement). The performance impact seems minor.
- added target_seq option for seq2seq models (outputs a shifted sequence instead of a single value)
- added prediction gap parameter
- added stateful option to ease parameters tuning for stateful models (experimental)
- added a host of assertions in init method

**Regression tests**
- adapted previous regression tests to the new length semantic (length=5 and sample_rate=2 => windows size = 10)
- added several new regression tests


TanguyUrvoy:
Checked code for pep8 compliance, and python 2/3 compatibility

TanguyUrvoy:
- autoprep8
- python 2/3 compatibility

PR #258, opened 2018-05-25
Fix NASNet Weights

titu1994:
Implements the corrections in Keras master (https://github.com/keras-team/keras/pull/10209) to properly generate the model and load weights.

### DIFF
--- a/keras_contrib/applications/nasnet.py
+++ b/keras_contrib/applications/nasnet.py
@@ -23,6 +23,7 @@ from keras.models import Model
 from keras.layers import Input
 from keras.layers import Activation
 from keras.layers import Dense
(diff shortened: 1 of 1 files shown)


titu1994:
@ahundt All tests failing for some reason.

ahundt:
yeah there was a change in upstream keras we need to account for. I started fixing it in https://github.com/keras-team/keras-contrib/pull/256 but apparently the test needs to be updated too

titu1994:
It is becoming a hassle to run tests against master Keras. Perhaps the partition of Keras into subcomponents will help with that.

titu1994:
@ahundt Thanks for the review !

PR #262, opened 2018-05-29
Added fix for dead ReLU detector with Activation layer

arodiss:
fixes https://github.com/keras-team/keras-contrib/issues/261


arodiss:
It's done.
Tests are in the same state as they were before this PR

titu1994:
LGTM, if tests pass then will merge.

PR #256, opened 2018-05-23
remove legacy support to conform with keras master

ahundt:
https://github.com/keras-team/keras/commit/a637960fab61b66848a36e6a5caf0204c155af01
https://github.com/keras-team/keras/pull/10077

PR #263, opened 2018-05-29
remove legacy support to conform with keras master

ahundt:
Updating keras-contrib to no longer use the models legacy code, since it has been moved or removed. This means legacy (1.x?) models will no longer load and save with load_all_weights and save_all_weights.

https://github.com/keras-team/keras/commit/a637960fab61b66848a36e6a5caf0204c155af01
https://github.com/keras-team/keras/pull/10077

replaces #256

### DIFF
--- a/keras_contrib/utils/save_load_utils.py
+++ b/keras_contrib/utils/save_load_utils.py
@@ -28,10 +28,7 @@ def save_all_weights(model, filepath, include_optimizer=True):
 
     with h5py.File(filepath, 'w') as f:
         model_weights_group = f.create_group('model_weights')
-        if legacy_models.needs_legacy_support(model):
-            model_layers = legacy_models.legacy_sequential_layers(model)
-        else:
-            model_layers = model.layers
+        model_layers = model.layers
         saving.save_weights_to_hdf5_group(model_weights_group, model_layers)
 
         if include_optimizer and hasattr(model, 'optimizer') and model.optimizer:
--- a/keras_contrib/utils/save_load_utils.py
+++ b/keras_contrib/utils/save_load_utils.py
@@ -4,7 +4,6 @@ import h5py
 import keras.backend as K
 from keras import optimizers
 from keras.engine import saving
-from keras.legacy import models as legacy_models
 
 
 def save_all_weights(model, filepath, include_optimizer=True):


ahundt:
wow its failing on `import tensorflow as tf`... fun!

ahundt:
Well I've been periodically doing a "manual merge" when it breaks. Starting over with the keras config and adding a couple lines for keras-contrib might be an option though. I noted this sort of issue in keras-team/keras#9256 since they're proposing that many more repos like this one be created in the future. I guess then we could rely on whatever infrastructure they create to support that.

titu1994:
@ahundt I've been following that thread as well for some time. I've been reviewing some large frameworks to see if we could move forward similarly, but till date I've not found many large scale projects that work in fragments. 

In any case, constant testing against master is quite brittle. I was considering proposing an eventual consistency behaviour where we could parallel the versioning of the fragments, and the fragments only test against the latest tag of Keras proper, not it's master branch. 

When Keras updates, the fragments update to test against the new release, and work backwards from Keras commits to fix tests. Once all are fixed, the fragments also releases a tag to match Keras release. 

This still involves significant manual work, especially the back tracking of breaking changes in Keras internals, however it ensures at least 1 tag of the fragment that is available to users. 

Currently, if you try to download contrib from github, you are left wondering if the tests are passing, whether the package is versioned, and this is useless for most research tasks for replication.

ahundt:
@titu1994 I agree, but I've discovered that you can only put in your two cents and since the lead developer is putting in the time to maintain the project any changes are their decision to make. Often decisions like this are reverted after a year or two (homebrew), but there are some projects like Qt and Boost where they keep all the projects separate but create a big bundled release periodically. I'm not aware of many truly successful splintered APIs. I agree it isn't ideal for easing reproducible research and there will be a few other problems like I mentioned in the other thread. That said, it will at worst just be a manageable nuisance in practice, so it is not worth making strong and pointed objections. Strong objections would be unlikely to change the course of the project and might sour more important future discussions.

Your suggestion for keras-contrib simply tracking the latest release and having someone make pull requests to update to new future releases sounds like a more reliable approach to keeping keras-contrib builds passing. Could you create a new issue containing a description of such a change so there could be a discussion period?

titu1994:
@ahundt I agree completely to your first point. I did not mean to sound as if it was a complaint, merely raising a discussion point. 

In any case, the codebase was successfully split (as I believe the preprocessing module has been turned into it's own project). Hopefully, contrib can benefit further from this. 

As to my earlier proposal, I think while distributed databases benefit from it, in practice, it will be difficult to disentangle two repositories which depend on each other. Data after all is mostly independent of other data, code rarely is, even with best design practices. 

A simple case of failire in such a eventual consistency paradigm is the moment following an update from Keras which breaks earlier usage of submodule A. 

Keras version B will depend on updates to submodule A, yet submodule A has not pushed out a release to match Keras version B requirements yet. For that time period, Keras tests will fail, and people will need to pip install from the master branch of submodule A to even use Keras version B. 

This can still be handled by using synchronized update schedules, but it adds significant scheduling overhead and is not reasonable. 

In the worst case, submodule A may be caught off guard and simply not have a patch prepared. I'm which case Keras will have to revert until a patch is prepared. 

In my opinion, let's see how the current strategy will handle a few updates accross the submodules and Keras. I don't think there will be significant issues with the current strategy.

titu1994:
@ahundt Should we try integrating the current travis config with the config from keras master and see if that passes the tests?

ahundt:
I believe this, or something equivalent has been merged in https://github.com/keras-team/keras-contrib/pull/274

PR #264, opened 2018-05-30
 Keras inference time optimizer

ZFTurbo:
Reduce neural net structure (Conv + BN -> Conv)
Also works:
DepthwiseConv2D + BN -> DepthwiseConv2D
SeparableConv2D + BN -> SeparableConv2D

This code takes on input trained Keras model and optimize layer structure and weights in such a way that model became much faster (~30%), but works identically to initial model. It can be extremely useful in case you need to process large amount of images with trained model. Reduce operation was tested on all Keras models zoo. See comparison table and full description by link:
https://github.com/ZFTurbo/Keras-inference-time-optimizer

### DIFF
--- a/keras_contrib/utils/kito.py
+++ b/keras_contrib/utils/kito.py
@@ -1,0 +1,311 @@
+# coding: utf-8
+__author__ = 'Roman Solovyev (ZFTurbo), IPPM RAS: https://github.com/ZFTurbo/'
+
+'''
+Reduce neural net structure (Conv + BN -> Conv)
+Also works:
+DepthwiseConv2D + BN -> DepthwiseConv2D
+SeparableConv2D + BN -> SeparableConv2D
+
+This code takes on input trained Keras model and optimize layer structure and weights in such a way
+that model became much faster (~30%), but works identically to initial model. It can be extremely
+useful in case you need to process large amount of images with trained model. Reduce operation was
+tested on all Keras models zoo. See comparison table and full description by link:
+https://github.com/ZFTurbo/Keras-inference-time-optimizer
+'''
+
+DEBUG = False
+import numpy as np
+
+
+def get_input_layers_ids(model, layer):
+    res = dict()
+    for i, l in enumerate(model.layers):
+        layer_id = str(id(l))
+        res[layer_id] = i
+
+    inbound_layers = []
+    layer_id = str(id(layer))
+    for i, node in enumerate(layer._inbound_nodes):
+        node_key = layer.name + '_ib-' + str(i)
+        if node_key in model._container_nodes:
+            for inbound_layer in node.inbound_layers:
+                inbound_layer_id = str(id(inbound_layer))
+                inbound_layers.append(res[inbound_layer_id])
+    return inbound_layers
+
+
+def get_output_layers_ids(model, layer):
+    res = dict()
+    for i, l in enumerate(model.layers):
+        layer_id = str(id(l))
+        res[layer_id] = i
+
+    outbound_layers = []
+    layer_id = str(id(layer))
+    for i, node in enumerate(layer._outbound_nodes):
+        node_key = layer.name + '_ib-' + str(i)
+        if node_key in model._container_nodes:
+            outbound_layer_id = str(id(node.outbound_layer))
+            outbound_layers.append(res[outbound_layer_id])
+    return outbound_layers
+
+
+def get_copy_of_layer(layer):
+    from keras.applications.mobilenet import relu6
+    from keras.layers.core import Activation
+    from keras import layers
+    config = layer.get_config()
+
+    # Non-standard relu6 layer (from MobileNet)
+    if layer.__class__.__name__ == 'Activation':
+        if config['activation'] == 'relu6':
+            layer_copy = Activation(relu6, name=layer.name)
+            return layer_copy
+
+    # DeepLabV3+ non-standard layer
+    if layer.__class__.__name__ == 'BilinearUpsampling':
+        from neural_nets.deeplab_v3_plus_model import BilinearUpsampling
+        layer_copy = BilinearUpsampling(upsampling=config['upsampling'], output_size=config['output_size'], name=layer.name)
+        return layer_copy
+
+    layer_copy = layers.deserialize({'class_name': layer.__class__.__name__, 'config': config})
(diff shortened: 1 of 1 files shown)


ahundt:
The style here doesn't seem to match that of other keras code. Could you fix that?

titu1994:
Are there no tests other than the benchmark test from your repo that can be performed ? A benchmark test isn't what would be feasible inside Travis, especially with such large models.

Once the style fixes are finished, this can be merged if it includes at least some form of test and all tests pass.

ZFTurbo:
@titu1994 I added "mobilenet_small" benchmark. It run several seconds on my machine. I think it can be used as test for kito.py code. 

Should I include tests directly in kito.py file?

@ahundt: Regarding style. I saw other files in "utils" folder and didn't see any specific style pattern. Could you please clarify what exactly should I change? Do you mean comments in functions or something else?

ahundt:
tests should go in the tests folder, take a look how those are written.

titu1994:
@ZFTurbo The test is short on Tensorflow, but I doubt its fast on Theano. In any case, it's better than no test at all.

As @ahundt mentioned, tests should go into the `tests` directory, specifically `tests/keras_contrib/utils/*_test.py` where you can replace * with some name. For the format, please see other tests or tests from core Keras.

ZFTurbo:
I added test.

ahundt:
@titu1994 @ZFTurbo this looks pretty good to me now, think it should be merged?

ZFTurbo:
@ahundt I'm totally fine with merge. )

keras_contrib/utils/kito.py
```diff
@@ -0,0 +1,309 @@
...
+            if get_keras_sub_version() == 1:
+                from keras.applications.mobilenet import relu6
+            else:
+                from keras_applications.mobilenet import relu6
+            layer_copy = Activation(relu6, name=layer.name)
+            return layer_copy
+
+    # DeepLabV3+ non-standard layer
+    if layer.__class__.__name__ == 'BilinearUpsampling':
+        from neural_nets.deeplab_v3_plus_model import BilinearUpsampling
```
ahundt: Just noticed this, there is now a bilinear upsampling layer in keras itself.
https://github.com/keras-team/keras/pull/10994

tests/keras_contrib/utils/kito_test.py
```diff
@@ -0,0 +1,121 @@
...
+        print(model.summary())
+        start_time = time.time()
+        model_reduced = reduce_keras_model(model, verbose=True)
+        print("Reduction time: {:.2f} seconds".format(time.time() - start_time))
+        print(model_reduced.summary())
+        print('Initial model number layers: {}'.format(len(model.layers)))
+        print('Reduced model number layers: {}'.format(len(model_reduced.layers)))
+        print('Compare models...')
+        if model_name in ['nasnetlarge', 'deeplab_v3plus_mobile', 'deeplab_v3plus_xception']:
+            max_error = compare_two_models_results(model, model_reduced, test_number=10000, max_batch=128)
```
ahundt: how long does this test take to run?

keras_contrib/utils/kito.py
```diff
@@ -0,0 +1,309 @@
...
+        if node_key in network_nodes:
+            outbound_layer_id = str(id(node.outbound_layer))
+            if outbound_layer_id in res:
+                outbound_layers.append(res[outbound_layer_id])
+            else:
+                print('Warning, some problem with outbound node on layer {}!'.format(layer.name))
+    return outbound_layers
+
+
+def get_copy_of_layer(layer, verbose=False):
```
ahundt: how do users do this for their own non-standard layers?

keras_contrib/utils/kito.py
ZFTurbo: It has other name in Keras. This code made specially for DeepLab V3+. It still uses it's own implementation.

tests/keras_contrib/utils/kito_test.py
ZFTurbo: Several seconds

keras_contrib/utils/kito.py
ZFTurbo: They need to change the code ( Like here:

    # DeepLabV3+ non-standard layer
    if layer.__class__.__name__ == 'BilinearUpsampling':
        from neural_nets.deeplab_v3_plus_model import BilinearUpsampling

I'm not sure how to make it universal, since parameters are different.

keras_contrib/utils/kito.py
bhack: /cc @bonlime

ahundt: We can't have specialized code for things not included in keras-contrib. You'll need to come up with a way to parameterize this, such as by passing functions, or a dictionary.

ZFTurbo:
1) Added more tests
2) Removed non-standard layers. Probably need to add custom_objects parameter. Will do it later.
3) Added support for layers of type "Model".

keras_contrib/utils/kito.py
ahundt: You can also have a look at keras itself registers functions. If you look at the python file for loss, at the very bottom there is some registration code.

gabrieldemarmiesse:
Thank you @ZFTurbo for your work.

I've looked at the result presented in your page and the number are impressive.
Since you're using tensorflow, I might add that there is XLA (maybe you already know about it) which does a similar job.

If it's not too much trouble, can I ask you to run your benchmarks with XLA enabled?

Also as a side note. I started using keras for the first time with your scripts on kaggle a few years ago. Thanks for sharing! (remember the fish competition?) :)

ZFTurbo:
> 
> 
> Thank you @ZFTurbo for your work.
> 
> I've looked at the result presented in your page and the number are impressive.
> Since you're using tensorflow, I might add that there is XLA (maybe you already know about it) which does a similar job.
> 
> If it's not too much trouble, can I ask you to run your benchmarks with XLA enabled?

I heard about XLA but not tried it yet. Probably XLA will make KITO useless. I'll try it and post result here.

> Also as a side note. I started using keras for the first time with your scripts on kaggle a few years ago. Thanks for sharing! (remember the fish competition?) :)

Good to know it )) That's bad, but looks like PyTorch becoming mainstream now, instead of Keras.

PR #277, opened 2018-06-23
the location of _obtain_input_shape changed

xyp8023:
When I tried to import 'densenet' from 'keras_contrib.applications', there is an import error saying that could not import '_obtain_input_shape' from 'keras.applications.imagenet_utils'. 

Then I found out it is in the location of 'keras_applications.imagenet_utils' and if I just change the 'from keras.applications.imagenet_utils import _obtain_input_shape' to 'from keras_applications.imagenet_utils import _obtain_input_shape', it would work.


titu1994:
Interesting, this should not be the case normally, since keras dynamically adds keras_application modules to itself.

I guess the same cannot be said for the internal functions like `_obtain_input_shape`.

titu1994:
Thanks for the PR !

PR #284, opened 2018-07-13
Keras app

wangtong321:
keras_applications-->keras.applications


ahundt:
you may be using an old version of keras, these changes are correct to my knowledge, see https://github.com/keras-team/keras-applications

PR #287, opened 2018-07-18
[BatchRenorm] update running variance correctly

mrlzla:
Update moving running variance with batch running variance since it will work incorrect if we use large epsilon.

### DIFF
--- a/keras_contrib/layers/normalization.py
+++ b/keras_contrib/layers/normalization.py
@@ -316,7 +316,7 @@ class BatchRenormalization(Layer):
 
         # explicit update to moving mean and standard deviation
         self.add_update([K.moving_average_update(self.running_mean, mean_batch, self.momentum),
-                         K.moving_average_update(self.running_variance, std_batch ** 2, self.momentum)], inputs)
+                         K.moving_average_update(self.running_variance, var_batch, self.momentum)], inputs)
 
         # update r_max and d_max
         r_val = self.r_max_value / (1 + (self.r_max_value - 1) * K.exp(-self.t))


titu1994:
Currently, the BatchRenorm implementation follows the paper closely. 

A large epsilon, (I am assuming closer to 1), is rarely every used in any normalization procedure. Perhaps only for demonstration purposes that they are equivalent under certain circumstances. 

Can you expand your case of why this change is necessary?

mrlzla:
I pretty sure that is unnecessary, anyway there are some misunderstandings when update some metric related to variance with variable related to standart deviation.

mrlzla:
After each minibatch we have got running_variance metric, which is greater than actual value by  eps * (1 - momentum). Let eps=1e-3, momentum=0.99. Then we have an error 1e-5. Through 1000 minibatches we have got an error that is greater than actual value by 1e-2, that is very big I think.

titu1994:
Two things, can you demonstrate that this error affects training significantly, such that this fix is sufficient to correct. 

The error, if it exists, is not additive, and ann error of 1e-5 as you suggest may occur but it will be on a per batch basis, not add up to 1e-2 over all runs. 

Secondly, the source of the error in your opinion comes from the squaring of the standard deviation ? That error would be far smaller than the 1e-5 you suggest I think. 

In any case, I'm not averse to the change, but I want to make sure it doesn't reduce performance by mistake. If you could, please train a model with low batch sizes, and plot the loss graph. If it's similar to the plots in the discussion of the original inplementation of Batch Renorm, I'll merge it.

mrlzla:
Unfortunately I have no enough free resources to train some model with batch renorm and this pull request only based on mathematical non-conformity. So I cannot say if it impoves results or not.

titu1994:
Good analysis, but plugging in the values from your computation we find : 

> eps * (1 - momentum) as the error. 
> Xn_hat = Xn + err / (1 - momentum).

Xn_hat = Xn + (eps * (1-momentum)) / (1-momentum)

Therefore, Xn_hat = Xn + eps. I fail to see how you arrived at 1e-1 as your final error value. 

Since epsilon is normally 1e-3, the difference is small enough to be ignored.

mrlzla:
Yes, my fault. 
1e-3 is really small enough. I have no arguments more :)

Issue #291, opened 2018-07-30
Can't import keras-contrib with keras2.2.1+


keisen:
```
$ python
Python 3.5.5 (default, Jul 11 2018, 05:47:58) 
[GCC 7.2.1 20170915 (Red Hat 7.2.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import keras_contrib
/home/ec2-user/.pyenv/versions/3.5.5/lib/python3.5/importlib/_bootstrap.py:222: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
Using TensorFlow backend.
/home/ec2-user/.pyenv/versions/3.5.5/lib/python3.5/importlib/_bootstrap.py:222: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/home/ec2-user/.pyenv/versions/3.5.5/lib/python3.5/importlib/_bootstrap.py:222: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
/home/ec2-user/.pyenv/versions/3.5.5_test/lib/python3.5/site-packages/matplotlib/__init__.py:1067: UserWarning: Duplicate key in file "/home/ec2-user/.config/matplotlib/matplotlibrc", line #2
  (fname, cnt))
/home/ec2-user/.pyenv/versions/3.5.5_test/lib/python3.5/site-packages/matplotlib/__init__.py:1067: UserWarning: Duplicate key in file "/home/ec2-user/.config/matplotlib/matplotlibrc", line #3
  (fname, cnt))
/home/ec2-user/.pyenv/versions/3.5.5/lib/python3.5/importlib/_bootstrap.py:222: RuntimeWarning: numpy.dtype size changed, may indicate binary incompatibility. Expected 96, got 88
  return f(*args, **kwds)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ec2-user/.pyenv/versions/3.5.5_test/lib/python3.5/site-packages/keras_contrib/__init__.py", line 4, in <module>
    from . import layers
  File "/home/ec2-user/.pyenv/versions/3.5.5_test/lib/python3.5/site-packages/keras_contrib/layers/__init__.py", line 3, in <module>
    from .convolutional import *
[...]

PR #292, opened 2018-07-30
Fix: Can't import keras-contrib with keras2.2.1+ #291

keisen:
This PR is to fix #291 .

`normalize_data_format` function was moved to `keras.backend.common` from `keras.utils.conv_utils`  since keras 2.2.1 .
So, currently keras-contrib can't import with keras 2.2.1+ .

```
import keras_contrib  # ImportError !
```

This PR isn't thought compatibility.
(I believe that keras-contrib support only master branch of Keras.)
Also, Unit tests failed due to problems from before.

PR #293, opened 2018-07-30
Fix issue #291 - broken compatibility with Keras since 2.2.1 due to wrong import

wilderrodrigues:
Although there is another PR with this fix, it seems that it contains test errors and I also observed that the import is not right, as it is not importing from the Keras backend, as seen here:

```
https://github.com/keras-team/keras/blob/d40656e5799dfd22e96b8bab217638b2934a6894/tests/keras/utils/conv_utils_test.py

K.normalize_data_format('channels_middle')
```

This has been tested on a Databricks cluster and is working.

image

### DIFF
--- a/keras_contrib/layers/convolutional.py
+++ b/keras_contrib/layers/convolutional.py
@@ -12,7 +12,6 @@ from keras.engine import InputSpec
 from keras.layers.convolutional import Convolution3D
 from keras.utils.generic_utils import get_custom_objects
 from keras.utils.conv_utils import conv_output_length
-from keras.utils.conv_utils import normalize_data_format
 import numpy as np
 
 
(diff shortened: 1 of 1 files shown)


keisen:
Hi, @wilderrodrigues 

This PR is essentially as same as #292.
`keras.backend.normalize_data_format` is an alias for `keras.backend.common.normalize_data_format` in currently Keras.

https://github.com/keras-team/keras/blob/4db04d8701a64ee249dfe6ce09ebd89063bfb99c/keras/backend/tensorflow_backend.py#L22

It is confusing if there are multiple PRs that do the same fixes.

Thanks.

wilderrodrigues:
Hi @keisen,

Indeed, you are right.

I will close this PR and try to fix the issues with the tests on your instead.

The test could be failing to due a latest Numpy release. I compared the latest PR with green tests and mine/yours to see the difference. There are lots of complains coming from Numpy.

Cheers.

PR #301, opened 2018-09-01
Add backward compatibility to PR#292

keisen:
Hi, @ahundt .

This PR is the one of idea for your comment (i.e., https://github.com/keras-team/keras-contrib/pull/292#issuecomment-417859929 ).

### DIFF
--- a/keras_contrib/layers/convolutional.py
+++ b/keras_contrib/layers/convolutional.py
@@ -12,7 +12,11 @@ from keras.engine import InputSpec
 from keras.layers.convolutional import Convolution3D
 from keras.utils.generic_utils import get_custom_objects
 from keras.utils.conv_utils import conv_output_length
-from keras.backend.common import normalize_data_format
+import keras.utils.conv_utils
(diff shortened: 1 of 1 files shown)


ahundt:
Awesome! Can you add a comment explaining the release versions for which this is present so we can decide if/when to remove it in the future?

ahundt:
Oh I also think there are a couple other places in the code where an equivalent line would need to be applied before the merge would be useful, maybe try searching for "keras_" in the code base.

keisen:
Ah, I noticed, currently keras-vis is having a problem one more like `import keras_applications`.
It is `from keras.engine import saving`.

The `saving` module was added to keras at version 2.2 and its functions was moved from `keras.engine.topology`, while renamed.
So when I improve this problem, I have to add modification to the function in `keras_contrib/utils/save_load_utils.py` to adjust different of function names and module names,

What do you think, whether it should be improved or not?
In the keras-vis document, keras-vis has compatibility with currently keras, so I think, to close this PR is not bad.

PR #302, opened 2018-09-04
feature/Padam

MFreidank:
Hello everyone,

[Padam](https://arxiv.org/abs/1806.06763) is a new adaptive optimizer that was introduced very recently and promises to combine fast convergence speed (as with adaptive optimizers like Adam) with a more robust convergence to good optima (as has been observed for SGD with momentum). I ported [this pytorch reference code](https://github.com/thughost2/Padam/blob/master/Padam.py) to keras. 

Disclaimer: I am not affiliated in any way with any of the authors of the Padam paper and merely ported the code out of curiosity. I would like to share the code with a larger audience, because their results look quite promising and the paper was trending recently on arxiv sanity preserver. 

Thanks in advance for your comments.
Best,
MFreidank

### DIFF
--- a/keras_contrib/optimizers/padam.py
+++ b/keras_contrib/optimizers/padam.py
@@ -71,9 +71,7 @@ class Padam(Optimizer):
             self.updates.append(K.update(v, v_t))
 
             # Partial momentum adaption.
-            p_t = p - (lr_t * (m_t / (denom ** (self.partial * 2))))
-
-            new_p = p_t
+            new_p = p - (lr_t * (m_t / (denom ** (self.partial * 2))))
 
             # Apply constraints.
             if getattr(p, 'constraint', None) is not None:
--- a/keras_contrib/optimizers/padam.py
+++ b/keras_contrib/optimizers/padam.py
@@ -16,12 +16,19 @@ class Padam(Optimizer):
             amsgrad: boolean. Whether to apply the AMSGrad variant of this
                 algorithm from the paper "On the Convergence of Adam and
                 Beyond".
-            partial: float >=0. Parameter controlling partial momentum adaption.
+            partial: float, 0 <= partial <= 0.5 . Parameter controlling partial momentum adaption. For `partial=0`, this optimizer behaves like SGD, for `partial=0.5`
+            it behaves like AMSGrad.
 
         # References
             - [Closing the Generalization Gap of Adaptive Gradient Methods in Training Deep Neural Networks](https://arxiv.org/pdf/1806.06763.pdf)
 
         """
+        if partial < 0 or partial > 0.5:
+            raise ValueError(
+                "Padam: 'partial' must be a positive float with a maximum "
+                "value of `0.5`, since higher values will cause divergence "
+                "during training."
+            )
         super(Padam, self).__init__(**kwargs)
         with K.name_scope(self.__class__.__name__):
             self.iterations = K.variable(0, dtype='int64', name='iterations')
--- a/keras_contrib/optimizers/padam.py
+++ b/keras_contrib/optimizers/padam.py
@@ -88,6 +88,7 @@ class Padam(Optimizer):
                   'beta_2': float(K.get_value(self.beta_2)),
                   'decay': float(K.get_value(self.decay)),
                   'epsilon': self.epsilon,
-                  'amsgrad': self.amsgrad}
+                  'amsgrad': self.amsgrad,
+                  "partial": self.partial}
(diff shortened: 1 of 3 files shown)


titu1994:
From the paper, it suggests `partial` should be `0.125` rather than `0.25` here. Also, you need to add the `partial` parameter to the end of the `config` dictionary inside `get_config`, otherwise it will not be restored correctly.

MFreidank:
Hello, thanks for your comments.
Yeah, there appears to be a discrepancy between the paper and the default in their pytorch code. I guess it makes sense to follow the paper. Also, good catch as far as the config dictionary is concerned, I will fix that.

titu1994:
Ah I forgot to add that if you could, add tests inside the keras_contrib/tests/optimizers/. Here, if you could duplicate the Keras optimizer tests (just the ones based for Adam should do) and replace Adam with Padam. 

Right now, the tests are failing anyway, but it would be important to have them when thats fixed. You can run the tests for Padam locally to check if they are passing if you have the time.

keras_contrib/optimizers/padam.py
```diff
@@ -0,0 +1,94 @@
...
+        return self.updates
+
+    def get_config(self):
+        config = {'lr': float(K.get_value(self.lr)),
+                  'beta_1': float(K.get_value(self.beta_1)),
+                  'beta_2': float(K.get_value(self.beta_2)),
+                  'decay': float(K.get_value(self.decay)),
+                  'epsilon': self.epsilon,
+                  'amsgrad': self.amsgrad,
+                  "partial": self.partial}
```
titu1994: Use single quotes here for partial.

keras_contrib/optimizers/padam.py
```diff
@@ -0,0 +1,94 @@
...
+                denom = (K.sqrt(vhat_t) + self.epsilon)
+                self.updates.append(K.update(vhat, vhat_t))
+            else:
+                denom = (K.sqrt(v_t) + self.epsilon)
+
+            self.updates.append(K.update(m, m_t))
+            self.updates.append(K.update(v, v_t))
+
+            # Partial momentum adaption.
+            p_t = p - (lr_t * (m_t / (denom ** (self.partial * 2))))
```
titu1994: Replace with `new_p` directly here.

keras_contrib/optimizers/padam.py
```diff
@@ -0,0 +1,94 @@
...
+
+            new_p = p_t
```
titu1994: Remove this due to above change.

keras_contrib/optimizers/padam.py
```diff
@@ -0,0 +1,94 @@
...
+            amsgrad: boolean. Whether to apply the AMSGrad variant of this
+                algorithm from the paper "On the Convergence of Adam and
+                Beyond".
+            partial: float >=0. Parameter controlling partial momentum adaption.
+
+        # References
+            - [Closing the Generalization Gap of Adaptive Gradient Methods in Training Deep Neural Networks](https://arxiv.org/pdf/1806.06763.pdf)
+
+        """
+        super(Padam, self).__init__(**kwargs)
```
titu1994: Add a line to check that `partial` lies between 0 and 0.5, else raise a ValueError stating that it must be a positive float that can have a maximum value of 0.5, since anything higher will cause divergence during training. 

Better to have a fail fast behavior for something as pivotal as an optimizer.

keras_contrib/optimizers/padam.py
```diff
@@ -0,0 +1,94 @@
...
+        # Arguments
+            lr: float >= 0. Learning rate.
+            beta_1: float, 0 < beta < 1. Generally close to 1.
+            beta_2: float, 0 < beta < 1. Generally close to 1.
+            epsilon: float >= 0. Fuzz factor. If `None`, defaults to `K.epsilon()`.
+            decay: float >= 0. Learning rate decay over each update.
+            amsgrad: boolean. Whether to apply the AMSGrad variant of this
+                algorithm from the paper "On the Convergence of Adam and
+                Beyond".
+            partial: float >=0. Parameter controlling partial momentum adaption.
```
titu1994: Document here that it should be >= 0 and <= 0.5.  Also, add a line that states what the two extreme ranges exhibit - 0.0 exhibiting SGD behavior and 0.5 exhibiting AMSGrad behavior.

titu1994:
Great work on the PR !

MFreidank:
@titu1994 Thanks a lot for your help!

PR #310, opened 2018-10-05
TypeError: 'zip' object is not subscriptable in the 42 line

yuanxiaosc:
TypeError: 'zip' object is not subscriptable. So we can first transform zip objects into Iterable list objects.


gabrieldemarmiesse:
thanks for the PR!

PR #314, opened 2018-10-12
Fix bug in BatchRenormalization

DaniyarM:
Checking if broadcasting need in same way as in `BatchNormalization` layer. Otherwise it caused an error with `Keras 2.2.4`


rgreenblatt:
I think this change is the same as the change I made here https://github.com/keras-team/keras-contrib/pull/325. I didn't realize this addressed the same problem at first.

gabrieldemarmiesse:
Closing as #325 was merged.

PR #315, opened 2018-10-13
Fix Padam unbiased learning rate

crowsonkb:
The formula for computing Padam's unbiased learning rate `lr_t` is currently only valid when `partial` = 1/2, because it was copied unchanged from Adam/AMSGrad where `partial` is effectively always 1/2. This PR corrects it so it is valid for any value of `partial`.

### DIFF
--- a/keras_contrib/optimizers/padam.py
+++ b/keras_contrib/optimizers/padam.py
@@ -54,7 +54,7 @@ class Padam(Optimizer):
                                                   K.dtype(self.decay))))
 
         t = K.cast(self.iterations, K.floatx()) + 1
-        lr_t = lr * (K.sqrt(1. - K.pow(self.beta_2, t)) /
+        lr_t = lr * (K.pow(1. - K.pow(self.beta_2, t), self.partial) /
                      (1. - K.pow(self.beta_1, t)))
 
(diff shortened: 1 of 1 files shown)


MFreidank:
@crowsonkb I am not convinced that this is actually necessary. [The original implementation](https://github.com/uclaml/Padam/blob/e0e8432b65c4abadafcfe4b0f2c913f9b83515e7/Padam.py#L84) does not do this. In fact, it seems to me as if they explicitly [changed their implementation](https://github.com/uclaml/Padam/commit/0264b9d9bd04eeb6380abb88d7f7717491149f3a) to undo the changes in your PR.  Notice also, that we do use the `partial` term in our [final update](https://github.com/keras-team/keras-contrib/blob/df1137d325f4865b41aca548066a1987a49d81d5/keras_contrib/optimizers/padam.py#L85). Can you provide more details on whether and if so why you still think this is necessary?

gabrieldemarmiesse:
@MFreidank please ignore the bot's message, my bot is up and running and it was just triggered for this PR.

MFreidank:
@gabrieldemarmiesse 
This has been stale since my comment from three months ago, so I suggest closing it.

PR #272, opened 2018-06-17
Some modification on CRF

linxihui:
1. Separated `loss` and accuracy metrics from class instance, so that `keras.models.load_model` works with CRF.
2. Fixed problem with `bias=False`.

PR #318, opened 2018-10-19
Some modification on CRF (parallel PR)

lzfelix:
This PR continues the work in #272 and addresses some of the existing issues. I wasn't able to commit directly into that fork, so I've created another one. Please let me know if there's a way to make my commits into the existing PR.

It summary:

* the previous `from keras_contrib import *` statements were removed
* instead of raising an error if the user tries to use the methods from the `CRF` class to retrieve the loss and accuracy functions, a warning message is displayed and the proper function is returned, as suggested by @titu1994 
* it's possible to load models with CRF layers using the `custom_objects` parameter of `load_model`. Please refer to the example in `tests/keras_contrib/layers/crf_test.py`. Maybe we could work further in the layer implementation to remove the need to use `custom_objects` as well...


lzfelix:
It seems that the build failures are being caused by something from the master branch, rather than the PR commits.

gabrieldemarmiesse:
Thank you for the PR. I fixed the master branch. Things should work now. And sorry for the delay. 

Could you resolve the conflict by merging master into your branch. Also we now have a documentation. Could you make sure that your modifications are reported in the docs? Thanks. To build the documentation, head to https://github.com/keras-team/keras-contrib/tree/master/contrib-docs . It's very easy.

lzfelix:
Sorry for the delay, @gabrieldemarmiesse. I had some issues yesterday, but the changes are in place now. Basically, I only had to add the docs and it seems that these changes can be merged now. There are examples on how to load models that use the CRF layer both in `test_crf.py` and in the docs of the layer.

gabrieldemarmiesse:
Thanks a lot @lzfelix . Hopefully there will be very few issues and PRs about it :)

It's a first step to revive keras-contrib. I would like to have a online documentation and also put the package on pipy. 

But it won't be managable if I need to know all the features here in detail. It's simply not practical (I also have to take care of keras, which takes time). So thank you for helping out!

lzfelix:
Sure, glad to help!

PR #325, opened 2018-11-13
Fixed python2 range vs python3 range bug

rgreenblatt:
In python3 range doesn't make a list so the condition could never be true.


gabrieldemarmiesse:
Thanks for the PR!

PR #331, opened 2018-12-15
Keras contrib docs

### LINKED ISSUE #220: create keras-contrib documentation website
We would like help with generating a keras-team.github.io/keras-contrib equivalent to http://keras.io/.

gabrieldemarmiesse:
fixes #220 

This generate a documentation which have the same structure as the one from `keras-team/keras`. Some pages were removed. The code is taken from the keras repo. Hopefully, in the future, we'll be able to import all the functions from keras which generate the documentation.

There are some formatting issues, but this is due to some docstrings not being properly formatted. This is outside of the scope of this PR.

Note that if the keras `autogen.py` was in a separate repo (something like `keras-team/docs-generation`), this would be a lot easier to maintain.


tests/test_documentation.py
```diff
@@ -0,0 +1,160 @@
...
+from itertools import compress
+
+import pytest
+
+modules = ['keras.layers', 'keras.models', 'keras',
+           'keras.backend.tensorflow_backend', 'keras.engine',
+           'keras.wrappers', 'keras.utils',
+           'keras.callbacks', 'keras.activations',
+           'keras.losses', 'keras.models', 'keras.optimizers']
+accepted_name = ['from_config']
```
Dref360: We need to change those modules to match keras_contrib

gabrieldemarmiesse:
It finally works. 

There is still work to do to fix all the formatting issues within the documentation. Contributions welcome.

PR #342, opened 2018-12-18
Add Yogi optimizer

MarcoAndreaBuchmann:
### Summary

This PR adds the Yogi optimizer (see [1] below) which, according to its authors, leads to even better performance than Adam with similar theoretical guarantees on convergence. It was presented at NIPS 2018.

[1] [Adaptive Methods for Nonconvex Optimization](
           https://papers.nips.cc/paper/8186-adaptive-methods-for-nonconvex-optimization)

### Related Issues

(None)

### PR Overview

- [ y ] This PR requires new unit tests [y/n] (make sure tests are included)
- [ y ] This PR requires to update the documentation [y/n] (make sure the docs are up-to-date)
- [ y ] This PR is backwards compatible [y/n]
- [ n ] This PR changes the current API [y/n] (all API changes need to be approved by fchollet)

### DIFF
--- a/keras_contrib/optimizers/yogi.py
+++ b/keras_contrib/optimizers/yogi.py
@@ -19,7 +19,7 @@ class Yogi(Optimizer):
         - [Adaptive Methods for Nonconvex Optimization](
            https://papers.nips.cc/paper/8186-adaptive-methods-for-nonconvex-optimization)
 
-    If you open a issues or a pull request about the Yogi optimizer,
+    If you open an issue or a pull request about the Yogi optimizer,
     please add 'cc @MarcoAndreaBuchmann' to notify him.
     """
 
--- a/keras_contrib/optimizers/yogi.py
+++ b/keras_contrib/optimizers/yogi.py
@@ -87,4 +87,4 @@ class Yogi(Optimizer):
         return dict(list(base_config.items()) + list(config.items()))
 
 
-get_custom_objects().update({'yogi': Yogi})
+get_custom_objects().update({'Yogi': Yogi})
--- a/keras_contrib/optimizers/yogi.py
+++ b/keras_contrib/optimizers/yogi.py
@@ -1,7 +1,6 @@
 from keras import backend as K
 from keras.optimizers import Optimizer
 from keras.utils.generic_utils import get_custom_objects
-from keras.legacy import interfaces
 
 
 class Yogi(Optimizer):
@@ -42,7 +41,6 @@ class Yogi(Optimizer):
         self.epsilon = epsilon
         self.initial_decay = decay
 
-    @interfaces.legacy_get_updates_support
     def get_updates(self, loss, params):
         grads = self.get_gradients(loss, params)
         self.updates = [K.update_add(self.iterations, 1)]
--- a/keras_contrib/optimizers/yogi.py
+++ b/keras_contrib/optimizers/yogi.py
@@ -1,6 +1,7 @@
 from keras import backend as K
 from keras.optimizers import Optimizer
 from keras.utils.generic_utils import get_custom_objects
+from keras.legacy import interfaces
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
Thanks for the PR.

At the end of the docstring, can you add:

`If you open a issue or a pull request about the Yogi optimizer, please add 'cc @MarcoAndreaBuchmann' to notify him.`

I cannot be an expert in all the features which comes in this repo. Which means that you'll be notified if something is going on in github about it. This is to help out as it will not be easy for me to maintain everything. Would that work for you?

MarcoAndreaBuchmann:
Thanks for the quick feedback, @gabrieldemarmiesse ! I've added the requested line to the docstring so future contributors know whom to contact.

keras_contrib/optimizers/yogi.py
```diff
@@ -0,0 +1,92 @@
...
+    # References
+        - [Adaptive Methods for Nonconvex Optimization](
+           https://papers.nips.cc/paper/8186-adaptive-methods-for-nonconvex-optimization)
+
+    If you open a issues or a pull request about the Yogi optimizer,
+    please add 'cc @MarcoAndreaBuchmann' to notify him.
+    """
+
+    def __init__(self, lr=0.01, beta_1=0.9, beta_2=0.999,
+                 epsilon=1e-3, decay=0., **kwargs):
```
gabrieldemarmiesse: Is there a strong reason why epsilon can default to `K.epsilon()`? In keras, all optimizers have the default epsilon to `K.epsilon()`.

keras_contrib/optimizers/yogi.py
```diff
@@ -0,0 +1,92 @@
...
+        lr: float >= 0. Learning rate.
+        beta_1: float, 0 < beta < 1. Generally close to 1.
+        beta_2: float, 0 < beta < 1. Generally close to 1.
+        epsilon: float >= 0. Fuzz factor. If `None`, defaults to `K.epsilon()`.
+        decay: float >= 0. Learning rate decay over each update.
+    # References
+        - [Adaptive Methods for Nonconvex Optimization](
+           https://papers.nips.cc/paper/8186-adaptive-methods-for-nonconvex-optimization)
+
+    If you open a issues or a pull request about the Yogi optimizer,
```
gabrieldemarmiesse: an issue or a pull request

keras_contrib/optimizers/yogi.py
MarcoAndreaBuchmann: Thanks for having such a close look at the implementation.

In the [paper](https://papers.nips.cc/paper/8186-adaptive-methods-for-nonconvex-optimization), `epsilon` is `1e-2` (Tab 1) or `1e-3` (Tab 2), whereas the backend's `epsilon` is orders of magnitude smaller (e.g.`1e-7` for [tensorflow](https://www.tensorflow.org/api_docs/python/tf/keras/backend/epsilon)). In my opinion, providing this default value is therefore necessary, as not doing so would mean that users would have a parameter that's off by orders of magnitude from the values in the paper it's meant to reproduce.

I understand your point of preferring consistency across optimizers, but in this case it would not lead to a faithful reproduction of the paper.

keras_contrib/optimizers/yogi.py
MarcoAndreaBuchmann: Thanks for spotting that, seems I copy-pasted too enthusiastically. Will fix that in a minute!

keras_contrib/optimizers/yogi.py
gabrieldemarmiesse: Ok it makes sense. Let's keep it like that.

gabrieldemarmiesse:
Thanks for the pull request!

PR #373, opened 2018-12-27
LArs optimization

roya0045:
Lars implementation based on the tensorflow implementation.


gabrieldemarmiesse:
It seems that your code is not PEP8 complient. See the travis log for detail. Also you who uld include tests to check that your optimizer works. See the `tests` directory.

roya0045:
I just added the test function and cleaned up some of the code, I'll keep an eye on the travis build to ensure the code is pep8. Thanks for the feedback.

stante:
Your optimizer can't be correctly deserialized. The problem is that you return in get_config `lr` instead of `learning_rate`, `etaa` instead of `eeta` and `nesterov` instead of `use_nesterov`. When the class is desrialized this arguments are given along with the deserialized values.

To be harmonized with other existing optimizers I would suggest the named argument for the learning rate should be `lr` as it is with all other optimizers. Also `nesterov` I think would be more appropriate than `use_nesterov` because SGD also already uses `nesterov`.

roya0045:
@stante Thanks a lot! Yeah I'm juggling too many versions/implementations and it's showing.

roya0045:
I'm not sure what might be causing the error with Theano since I'm not familiar with this backend.

gabrieldemarmiesse:
Thano doesn't handle broadcasting automatically. I think this is where the issue comes from (At least this is what I think it is from the stacktrace).

roya0045:
@gabrieldemarmiesse From what I see theano should handle it properly, the other optimization algorithms have similar syntax for weight and velocity update. I think the issue might come from my condition.

For the condition to work I must have the same condition shape and output shape, which is the only noticeable difference.

roya0045:
It took me some time to get back to this and I didn't want to apply this fix but this is the simplest to solve the error.

roya0045:
@gabrieldemarmiesse Did you need anything else done before merging the branch?

keras_contrib/optimizers/lars.py
```diff
@@ -0,0 +1,110 @@
...
+            self.iterations = K.variable(0, dtype='int64', name='iterations')
+            self.lr = K.variable(lr, name='lr')
+            self.momentum = K.variable(momentum, name='momentum')
+            self.weight_decay = K.variable(weight_decay, name='weight_decay')
+            self.eeta = K.variable(eeta, name='eeta')
+        self.epsilon = epsilon
+        self.skip_list = skip_list
+        self.nesterov = nesterov
+
+    def get_config(self):
```
gabrieldemarmiesse: The get_config generally comes at the end. After the get_update (this is just style).

keras_contrib/optimizers/lars.py
```diff
@@ -0,0 +1,110 @@
...
+                new_p = p + v
+
+            # Apply constraints.
+            if getattr(p, 'constraint', None) is not None:
+                new_p = p.constraint(new_p)
+
+            self.updates.append(K.update(p, new_p))
+        return self.updates
+
+get_custom_objects().update({'LARS': LARS})
```
gabrieldemarmiesse: Please do not add this line. I know it's been done a lot in keras-contrib but this is going to change in the future. They're all going to be removed. Users will have to update custom objects manually.

tests/keras_contrib/optimizers/lars_test.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import print_function
+from keras_contrib.tests import optimizers
+from keras_contrib.optimizers import lars
+
+optimizers._test_optimizer(lars.LARS(0.01))
```
gabrieldemarmiesse: Please put it in functions called `test_lars` and `test_lars_nesterov`.

Can you also add a `test_lars_skip` where you create a model and use `skip_list`? The model doesn't need to be big. a model.fit on a single epoch should be enough.

keras_contrib/optimizers/lars.py
roya0045: What is the suggested way to prevent deserialization errors?

keras_contrib/optimizers/lars.py
```diff
@@ -0,0 +1,108 @@
...
+                 epsilon=0.0,
+                 skip_list=None,
+                 nesterov=False,
+                 **kwargs):
+
+        if momentum < 0.0:
+            raise ValueError("momentum should be positive: %s" % momentum)
+        if weight_decay < 0.0:
+            raise ValueError("weight_decay is not positive: %s" % weight_decay)
+        print(kwargs)
```
RaphaelMeudec: Not sure we want to keep this print

tests/keras_contrib/optimizers/lars_test.py
```diff
@@ -0,0 +1,29 @@
...
+   model.add(Dense(32, activation='relu', input_dim=30))
+   model.add(Dense(1, activation='sigmoid'))
+   model.compile(loss='mse', optimizer=lars.LARS(0.01,skip_list=['bias',]))
+   model.fit(X, y, epochs=3)
+  
+  
+
+  
+  
+test_base_lars()
```
RaphaelMeudec: pytest automatically runs all function that contains `test_`, so you don't need to add those calls in the file

tests/keras_contrib/optimizers/lars_test.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import print_function
+import numpy as np
+from keras_contrib.tests import optimizers
+from keras_contrib.optimizers import lars
+from keras.models import Sequential
+from keras.layers import Dense
+
+def test_base_lars():
```
RaphaelMeudec: To be ok with pep8 rules, you should have 2 blank lines between functions

keras_contrib/optimizers/lars.py
```diff
@@ -0,0 +1,108 @@
...
+                                 K.expand_dims((self.eeta * w_norm /
+                                                (g_norm + self.weight_decay * w_norm +
+                                                 self.epsilon)) * self.lr),
+                                 K.ones([1]) * self.lr)
+        if K.backend() == 'theano':
+            scaled_lr = scaled_lr[0]  # otherwise theano raise broadcasting error
+        # momentum
+        moments = [K.zeros(K.int_shape(p), dtype=K.dtype(p)) for p in params]
+        self.weights = [self.iterations] + moments
+        for p, g, m in zip(params, grads, moments):
```
RaphaelMeudec: You should use more explicit variable names like

```
param, grad, moment in zip(params, grads, moments):
    velocity = (m * self.momentum)
```

roya0045:
@gabrieldemarmiesse I might have to remove the skip_list option that was in the TF implementation. This variable was based on the assumptions that the parameters provided had a name. With keras is would seem that the parameters are passed as a list. The option would be to ask the used for an index to ignore or find a way to match the name to an index.

Otherwise I'll remove this parameter since it doesn't seem viable in keras, is that ok with you?

gabrieldemarmiesse:
I believe that variables have names in keras. It is impacted by the layer name though, so I'm not sure it's a good idea to rely on it. Can you detail which kind of parameter/variable is skipped when using this optimizer?

roya0045:
> I believe that variables have names in keras. It is impacted by the layer name though, so I'm not sure it's a good idea to rely on it. Can you detail which kind of parameter/variable is skipped when using this optimizer?

By default none but judging from the implementation this is based on, the goal would be to prevent rescaling of the bias and batch_normalization weights. but it seems that the whole rescaling part is skipped when one of the name is encountered. This is strange but I assume this can be safely ignored for now.

gabrieldemarmiesse:
@RaphaelMeudec if you have the time, could you take a final look at it one of these days?
@roya0045 please add lars to the `__init__.py` file in optimizers. It will allow users to do `from keras_contrib.optimizers import LARS`

gabrieldemarmiesse:
@roya0045 also know that you'll be tagged automatically when pull request or issues are opened about this feature.

roya0045:
@gabrieldemarmiesse Very well

RaphaelMeudec:
@roya0045 Your test is not passing because you've set an alias as `lars = LARS`, so when you do `from keras_contrib.optimizers import lars`, you're importing the class LARS and not the module, so you should  simply call `optimizers._test_optimizer(lars0.01))`

@gabrieldemarmiesse Apart from the test not passing, everything looks good to me

roya0045:
> @roya0045 Your test is not passing because you've set an alias as `lars = LARS`, so when you do `from keras_contrib.optimizers import lars`, you're importing the class LARS and not the module, so you should simply call `optimizers._test_optimizer(lars0.01))`
> 
> @gabrieldemarmiesse Apart from the test not passing, everything looks good to me

Yeah I was on the fence about using the capitalized version or not but since most acronyms are in lower case I went with the alias and forgot to update my tests. Thanks for the reminder!

RaphaelMeudec:
@roya0045 Tests are still not passing because what you have imported in `lars_test.py` is alread the optimizer. You should just call `optimizers._test_optimizer(lars(0.01))` instead of `optimizers._test_optimizer(lars.lars(0.01))`

RaphaelMeudec:
@gabrieldemarmiesse I think it can be merged now

@roya0045 Thanks for the contribution!

gabrieldemarmiesse:
Thanks @roya0045 and @RaphaelMeudec

PR #387, opened 2018-12-30
Making keras-contrib compatible with tf.keras

### LINKED ISSUE #376: Any plan to make the repo usable for tf.keras ?
Hi, 

Modules in keras-applications repo uses a strategy such that one has a choice to use either keras or tf.keras i.e. there are no imports of either of them (keras / tf.keras) in the code thereby not requiring an explicit dependency 

Is there any plan to use the same strategy for this repository as well ?

Regards & thanks
Kapil

gabrieldemarmiesse:
closes #376


convert_to_tf_keras.py
```diff
@@ -0,0 +1,56 @@
+import os
+
+list_conversions = [('import keras.', 'import tensorflow.keras.'),
```
farizrahman4u: Instead of doing text replace on the code base, cant we do this in a more programmatic/pythonic way? Just like how we handle multiple backends in keras?

Here is what I propose:

1 - We have a config.json in user/.keras-contrib which mentions whether to use keras or tf.keras
2 - We have a single place in the codebase which handles this duality . Lets call it `keras_lib.py`:

```python
keras_to_use = read('config.json')

if keras_to_use == "keras":
     module = "keras"
elif keras_to_use == "tf.keras":
     module = "tensorflow.keras"

globals().update(importlib.import_module(module).__dict__)
```

3 - Import keras_lib instead of keras everywhere else in the codebase
```python
from .keras_lib import Dense

class MyCustomDense(Dense):
    pass
```

4 - Profit!!

gabrieldemarmiesse: This is fine by me as long as it doesn't confuse too much the contributors.
Let's say I'm a contributor and I want to simply do:
```python
from keras.layers import Conv2D
import keras.backend as K
from keras.callbacks import TensorBoard
```

what would it look like with your proposition? It's not clear to me.

gabrieldemarmiesse: I might add that I'm not familiar with those python imports tricks so it's a bit hard for me to follow.

farizrahman4u:
```python
from keras_contrib.keras_lib.layers import Conv2D
from keras_contrib.keras_lib import backend as K
from keras_contrib.keras_lib.callbacks import TensorBoard
```

(You can also relative import keras_lib since its in the same package)

gabrieldemarmiesse: I'm a bit worried that contributors would read the code and think that they have to access the layers of keras contrib by doing `from keras_contrib.keras_lib.layers`.

Also when a contributor will read 
`from keras_contrib.keras_lib import backend as K` he/she will wonder what's going on.

If a normal reader enter a file in a codebase and can't even understand the first lines written, I believe it's going to slow down contributions. Readers will have to be aware of this hack to understand imports.

gabrieldemarmiesse: At first, I wanted a fork of the codebase for tf.keras, but I believe this fork can be generated automatically, also, this fork can be tested in travis, to ensure that we don't catch bugs too late.

farizrahman4u: Agree. It does look confusing. Lets do it your way,

gabrieldemarmiesse: Cool, I'll continue working on it. I hope to finish by this weekend.

gabrieldemarmiesse:
@farizrahman4u could you review this PR when/if you have the time? Thanks in advance.

CONTRIBUTING.md
```diff
@@ -61,6 +61,44 @@ We love pull requests. Here's a quick guide:
...
+```
+python convert_to_tf_keras.py
+```
+
+To convert your codebase back to keras-team/keras, do:
+```
+python convert_to_tf_keras.py --revert
+```
+
+Not that you are strongly encouraged to commit your code before in case the parsing would go wrong. To discard all the changes you made since the previous commit:
```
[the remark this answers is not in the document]
RaphaelMeudec: Are we sure we want to encourage users to use the `git reset --hard` ? Something like `git stash` would be less definitive.

convert_to_tf_keras.py
```diff
@@ -0,0 +1,59 @@
...
+import sys
+
+list_conversions = [('import keras.', 'import tensorflow.keras.'),
+                    ('import keras ', 'from tensorflow import keras '),
+                    ('import keras\n', 'from tensorflow import keras\n'),
+                    ('from keras.', 'from tensorflow.keras.'),
+                    ('from keras ', 'from tensorflow.keras ')]
+
+
+def replace_imports(file_path, revert):
```
RaphaelMeudec: Would be more prudent to add a test, to make sure the created file is as expected (like we do with generated docs)

CONTRIBUTING.md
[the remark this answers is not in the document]
gabrieldemarmiesse: Sure, the method I describe here is very brittle. This is because my knowledge of GIT is quite limited. Do you have a workflow in mind which would be cleaner and more robust?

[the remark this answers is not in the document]
RaphaelMeudec: I would encourage to do a soft reset with `git reset HEAD^` and then stash all the current untracked modifications with `git stash`. In this case, you can retrieve those untracked modifications with `git stash pop`

convert_to_tf_keras.py
gabrieldemarmiesse: We already run all the tests with the generated files. Do you have a specific test in mind?

gabrieldemarmiesse: Ho, I see what you mean. I'm not in favor of including tests at running time (except for catching user errors) because this will complexify the codebase. Separating the tests and the implementation allow for a better readability. And when tests fail, we know exactly which use case caused it. It's more extensible.

CONTRIBUTING.md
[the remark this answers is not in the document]
gabrieldemarmiesse: Thanks for the tip!

gabrieldemarmiesse:
I seems that tensorflow.keras.backend doesn't have local_conv_1d and local_conv2d.

convert_to_tf_keras.py
gabrieldemarmiesse: @RaphaelMeudec do you have any simple tests in mind?

RaphaelMeudec: @gabrieldemarmiesse What I had in mind is :

- Create an asset file in the test directory, and apply `replace_imports` to it. Make sure that there is an output file created, and that the return value of the function is approriate
- Apply in reverse the same function to make sure generated file matches the original file

CONTRIBUTING.md
```diff
@@ -95,7 +95,10 @@ python convert_to_tf_keras.py --revert
 
 Note that you are strongly encouraged to commit your code before in case the parsing would go wrong. To discard all the changes you made since the previous commit:
 ```
-git reset --hard
+# saves a copy of your current codebase in the git stash and comes back to the previous commit
+git reset && git stash
```
RaphaelMeudec: Using `git stash` is enough, you don't need the `git reset`  [doc](https://www.git-scm.com/docs/git-stash)

RaphaelMeudec:
@gabrieldemarmiesse Final note on your PR, also I explained how I imagined the tests for the transformation.

convert_to_tf_keras.py
gabrieldemarmiesse: I added a test in the same file (we don't have a choice, we can't import the function directly). And added the command to run it in travis.

CONTRIBUTING.md
gabrieldemarmiesse: Done

gabrieldemarmiesse:
Any more comments about this?

gabrieldemarmiesse:
I'll merge now. I'll open an issue in the tensorflow repo about the input of the build function.

gabrieldemarmiesse:
Opened https://github.com/tensorflow/tensorflow/issues/25790 to track the weird incompatibility. I hope it'll go away at some point.

PR #396, opened 2019-01-07
Add LearningRateMultiplier wrapper for optimizers

stante:
### Summary
Optimizer have a model global learning rate. This PR adds a wrapper, which can be used with existing optimizers to provide a facility to specify different learning rates per layers in a network. The per layer learning rate is specified as a factor, which is multiplied with the learning rate of the wrapped optimizer. This wrapper can be used in the following way:
```python
multipliers = {'dense_1': 0.5, 'dense_2': 0.4}
opt = LearningRateMultiplier(SGD, lr_multipliers=multipliers, lr=0.001, momentum=0.9)
```
The example wrappes SGD and specifies `lr` and `momentum` for it. The layer which contain the string `'dense_1'` has a multiplier of `0.5` and the layer which contains the string `dense_2` has the multiplier of `0.4`.

Different multipliers for kernel and bias can be specified with:
```python
multipliers = {'dense_1/kernel': 0.5, 'dense_1/bias': 0.1}
```
### Related Issues
There are issues regarding this topic in keras https://github.com/keras-team/keras/issues/11934, https://github.com/keras-team/keras/issues/7912 and partially https://github.com/keras-team/keras/issues/5920

### DIFF
--- a/keras_contrib/optimizers/lr_multiplier.py
+++ b/keras_contrib/optimizers/lr_multiplier.py
@@ -1,0 +1,65 @@
+from keras.optimizers import Optimizer
+from keras.utils import get_custom_objects
+
+
+class LearningRateMultiplier(Optimizer):
+    """Optimizer wrapper for per layer learning rate.
+
+    This wrapper is used to add per layer learning rates by
+    providing per layer factors which are multiplied with the
+    learning rate of the optimizer.
+
+    Note: This is a wrapper and does not implement any
+    optimization algorithm.
+
+    # Arguments
+        optimizer: An optimizer class to be wrapped.
+        lr_multipliers: Dictionary of the per layer factors. For
+            example `optimizer={'conv_1/kernel':0.5, 'conv_1/bias':0.1}`.
+            If for kernel and bias the same learning rate is used, the
+            user can specify `optimizer={'conv_1':0.5}`.
+        **kwargs: The arguments for instantiating the wrapped optimizer
+            class.
+    """
+    def __init__(self, optimizer, lr_multipliers=None, **kwargs):
+        self._class = optimizer
+        self._optimizer = optimizer(**kwargs)
+        self._lr_multipliers = lr_multipliers or {}
+
+    def _get_multiplier(self, param):
+        for k in self._lr_multipliers.keys():
+            if k in param.name:
+                return self._lr_multipliers[k]
+
+    def get_updates(self, loss, params):
+        mult_lr_params = {p: self._get_multiplier(p) for p in params if self._get_multiplier(p)}
+        base_lr_params = [p for p in params if self._get_multiplier(p) is None]
+
+        updates = []
+        base_lr = self._optimizer.lr
+        for param, multiplier in mult_lr_params.items():
+            self._optimizer.lr = base_lr * multiplier
+            updates.extend(self._optimizer.get_updates(loss, [param]))
+
+        self._optimizer.lr = base_lr
+        updates.extend(self._optimizer.get_updates(loss, base_lr_params))
+
+        return updates
+
+    def get_config(self):
+        config = {'optimizer': self._class,
+                  'lr_multipliers': self._lr_multipliers}
+        base_config = self._optimizer.get_config()
+        return dict(list(base_config.items()) + list(config.items()))
+
+    def __getattr__(self, name):
+        return getattr(self._optimizer, name)
+
+    def __setattr__(self, name, value):
+        if name.startswith('_'):
+            super().__setattr__(name, value)
+        else:
+            self._optimizer.__setattr__(name, value)
+
+
+get_custom_objects().update({'LearningRateMultiplier': LearningRateMultiplier})
--- a/tests/keras_contrib/optimizers/lr_multiplier_test.py
+++ b/tests/keras_contrib/optimizers/lr_multiplier_test.py
@@ -1,0 +1,11 @@
+from keras_contrib.tests import optimizers
+from keras_contrib.optimizers import LearningRateMultiplier
+from keras.optimizers import SGD, Adam
+from keras.callbacks import LearningRateScheduler
+
+def test_lr_multiplier():
+    mult={'dense':10}
+    optimizers._test_optimizer(LearningRateMultiplier(SGD, lr=0.01, momentum=0.9, nesterov=True), target=0.95)
+    optimizers._test_optimizer(LearningRateMultiplier(SGD, lr_multipliers=mult, lr=0.001,
+                                                      momentum=0.9, nesterov=True), target=0.95)
+
--- a/keras_contrib/optimizers/__init__.py
+++ b/keras_contrib/optimizers/__init__.py
@@ -1,6 +1,7 @@
 from .ftml import FTML
 from .padam import Padam
 from .yogi import Yogi
+from .lr_multiplier import LearningRateMultiplier
 
 # aliases
 ftml = FTML


gabrieldemarmiesse:
It seems there is some pep8 errors and that the code isn't compatible with python 2 because of super() . Super takes two arguments in python 2. Usually it's the class  and `self`.

gabrieldemarmiesse:
You can find out more about the errors by looking at the travis logs.

keras_contrib/optimizers/lr_multiplier.py
```diff
@@ -0,0 +1,66 @@
...
+
+class LearningRateMultiplier(Optimizer):
+    """Optimizer wrapper for per layer learning rate.
+
+    This wrapper is used to add per layer learning rates by
+    providing per layer factors which are multiplied with the
+    learning rate of the optimizer.
+
+    Note: This is a wrapper and does not implement any
+    optimization algorithm.
```
gabrieldemarmiesse: What about two examples? 
* One where you specify manually the learning rate by using the strings as keys `{'conv_1/kernel':0.5, 'conv_1/bias':0.1}`
* One where you programmatically set the learning rates by iterating through the layers of the model (for big models this is useful). I suppose that it should be possible with a for loop and getting the `layer.name` as the key of the dictionary.

keras_contrib/optimizers/lr_multiplier.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Arguments
+        optimizer: An optimizer class to be wrapped.
+        lr_multipliers: Dictionary of the per layer factors. For
+            example `optimizer={'conv_1/kernel':0.5, 'conv_1/bias':0.1}`.
```
gabrieldemarmiesse: Typo: the keyword is lr_multipliers.

keras_contrib/optimizers/lr_multiplier.py
```diff
@@ -0,0 +1,66 @@
...
+
+    This wrapper is used to add per layer learning rates by
+    providing per layer factors which are multiplied with the
+    learning rate of the optimizer.
+
+    Note: This is a wrapper and does not implement any
+    optimization algorithm.
+
+    # Arguments
+        optimizer: An optimizer class to be wrapped.
```
gabrieldemarmiesse: I think optimizer should be an optimizer instance, not an optimizer class. Let's minimize the hackyness.

keras_contrib/optimizers/lr_multiplier.py
```diff
@@ -0,0 +1,66 @@
...
+        lr_multipliers: Dictionary of the per layer factors. For
+            example `optimizer={'conv_1/kernel':0.5, 'conv_1/bias':0.1}`.
+            If for kernel and bias the same learning rate is used, the
+            user can specify `optimizer={'conv_1':0.5}`.
+        **kwargs: The arguments for instantiating the wrapped optimizer
+            class.
+    """
+    def __init__(self, optimizer, lr_multipliers=None, **kwargs):
+        self._class = optimizer
```
gabrieldemarmiesse: I don't think underscores are needed.

tests/keras_contrib/optimizers/lr_multiplier_test.py
```diff
@@ -0,0 +1,14 @@
...
+from keras.optimizers import SGD, Adam
+from keras.callbacks import LearningRateScheduler
+
+
+def test_lr_multiplier():
+    opt1 = LearningRateMultiplier(SGD, lr=0.01, momentum=0.9, nesterov=True)
+    optimizers._test_optimizer(opt1, target=0.95)
+
+    mult = {'dense': 10}
+    opt2 = LearningRateMultiplier(SGD, lr_multipliers=mult,
```
gabrieldemarmiesse: Can you make a second function `test_lr_multiplier_layerwise` for this?

tests/keras_contrib/optimizers/lr_multiplier_test.py
```diff
@@ -0,0 +1,14 @@
...
+                                  lr=0.001, momentum=0.9, nesterov=True)
+    optimizers._test_optimizer(opt2, target=0.95)
```
gabrieldemarmiesse: We'll also need a third test `test_lr_multiplier_weightwise` where you use the format `{'layer_name/weight_name': lr}` to ensure that all configuration work.

And a fourth test with a more complex optimizer (ADAM would be a good fit)

tests/keras_contrib/optimizers/lr_multiplier_test.py
```diff
@@ -0,0 +1,14 @@
+from keras_contrib.tests import optimizers
+from keras_contrib.optimizers import LearningRateMultiplier
+from keras.optimizers import SGD, Adam
+from keras.callbacks import LearningRateScheduler
```
gabrieldemarmiesse: Unused import

keras_contrib/optimizers/lr_multiplier.py
```diff
@@ -0,0 +1,66 @@
...
+        return dict(list(base_config.items()) + list(config.items()))
+
+    def __getattr__(self, name):
+        return getattr(self._optimizer, name)
+
+    def __setattr__(self, name, value):
+        if name.startswith('_'):
+            super(LearningRateMultiplier, self).__setattr__(name, value)
+        else:
+            self._optimizer.__setattr__(name, value)
```
gabrieldemarmiesse: I don't think `__setattr__` and `__getattr__` are needed. By calling the right `super()` functions at the right places, everything should work. Ask me if you have any issues while removing them.

You'll likely have to have a `lr` parameter which will be the same as `self.optimizer.lr` since many callbacks expect a `lr` attribute.

keras_contrib/optimizers/lr_multiplier.py
```diff
@@ -0,0 +1,66 @@
...
+            If for kernel and bias the same learning rate is used, the
+            user can specify `optimizer={'conv_1':0.5}`.
+        **kwargs: The arguments for instantiating the wrapped optimizer
+            class.
+    """
+    def __init__(self, optimizer, lr_multipliers=None, **kwargs):
+        self._class = optimizer
+        self._optimizer = optimizer(**kwargs)
+        self._lr_multipliers = lr_multipliers or {}
+
```
gabrieldemarmiesse: You should call `super()` at the end of the `__init__` function. You can take a look at the source code of keras optimizers to see what happends.

keras_contrib/optimizers/lr_multiplier.py
```diff
@@ -0,0 +1,66 @@
...
+            self._optimizer.lr = base_lr * multiplier
+            updates.extend(self._optimizer.get_updates(loss, [param]))
+
+        self._optimizer.lr = base_lr
+        updates.extend(self._optimizer.get_updates(loss, base_lr_params))
+
+        return updates
+
+    def get_config(self):
+        config = {'optimizer': self._class,
```
gabrieldemarmiesse: Since optimizer will be an instance of the class optimizer, you should use the function `serialize_keras_object` which will serialize the optimizer for you.

Dicksonchin93:
will there be updates on this? if not can I make a new PR that adds this class to keras-contrib? @gabrieldemarmiesse @stante , will be enabling DiscriminativeLearningRate in general but not specifically only learning rate multiplier. 

I propose three settings, automatic learning rate decaying (cosine) from the base learning rate of the wrapped optimizer by layer, automatic learning rate decaying (cosine) from the base learning rate of the wrapped optimizer by convolutional blocks/groups, and this learning rate multiplier

gabrieldemarmiesse:
Keras contrib is currently deprecated. Please redicted the PRs to tensorflow/addons. It would be really nice if you could add that @Dicksonchin93 , a lot of people are asking for this feature :)

Dicksonchin93:
@gabrieldemarmiesse is there a reason why we shouldn't add this into keras directly?

gabrieldemarmiesse:
This was proposed a while back and rejected. The reason is that not enough
people use it to justify an API change of Keras. It's also not clear that
it's a best practice. Tensorflow addons was made exactly for this kind of
feature.

PR #425, opened 2019-02-04
Removing the usage of get_custom_objects.

gabrieldemarmiesse:
**- What I did**

I removed the modifications of global objects at import time.

**- How I did it**
removed the stuff and updated the readme.

**- How you can verify it**

Now the new test is using custom objects. Both optimizers and layers have tests where they are put in json and loaded back.


gabrieldemarmiesse:
Thanks @RaphaelMeudec !

wilderrodrigues:
Good I found it. I'm facing an issue now with loading the models in the old fashion way.

PR #432, opened 2019-02-06
Add missing parameters to get_config in Capsule Layer

SriRangaTarun:
**- What I did**
Added regularizer, initializer and constraint parameters to the get_config function in the Capsule layer.

**- How I did it**

**- How you can verify it**

---


SriRangaTarun:
@gabrieldemarmiesse I have fixed a little mistake which wasn't spotted.

gabrieldemarmiesse:
Thanks for noticing that. I wonder if there is a way to catch this kind of bugs automatically in unit tests.

PR #433, opened 2019-02-06
Add 'squash' (string) activation option in Capsule and update docs

SriRangaTarun:
**- What I did**
Added the squash activation function in activations/squash.py and added the option to use this activation function in the Capsule layer with a string argument. 

**- How I did it**

**- How you can verify it**

---


keras_contrib/layers/capsule.py
```diff
@@ -129,7 +122,8 @@ def __init__(self,
         self.routings = routings
         self.share_weights = share_weights
 
-        self.activation = activations.get(activation)
```
gabrieldemarmiesse: Do not modify this line. Users should do

```python
from keras_contrib.activations import squash
...
layer = Capsule(..., activation=squash)
```

keras_contrib/activations/squash.py
```diff
@@ -0,0 +1,7 @@
+def squash(x, axis=-1):
```
gabrieldemarmiesse: Please add a simple test for this function.

gabrieldemarmiesse:
Thanks for the PR!

PR #478, opened 2019-03-13
100% original Adam that allows to train x2 or x3 bigger networks.

iperov:
**- What I did**

added 100% original Adam optimizer with new option
```
tf_cpu_mode: only for tensorflow backend
0 - default, no changes.
1 - allows to train x2 bigger network on same VRAM consuming RAM
2 - allows to train x3 bigger network on same VRAM consuming RAM*2 and CPU power.
```

Batch size is very important parameter for GAN networks. So getting rid of optimizer's weights from VRAM, we can train higher batch size, sacrificing 10-20% of time per iteration.

**- How I did it**

accidentally discovered.

**- How you can verify it**

add `tf_cpu_mode=1` to Adam
and try x2 bigger network

### DIFF
--- a/keras_contrib/optimizers/adam.py
+++ b/keras_contrib/optimizers/adam.py
@@ -1,0 +1,106 @@
+from keras import backend as K
+from keras.optimizers import Optimizer
+
+class Adam(Optimizer):
+    """Adam optimizer.
+
+    Default parameters follow those provided in the original paper.
+
+    # Arguments
+        lr: float >= 0. Learning rate.
+        beta_1: float, 0 < beta < 1. Generally close to 1.
+        beta_2: float, 0 < beta < 1. Generally close to 1.
+        epsilon: float >= 0. Fuzz factor. If `None`, defaults to `K.epsilon()`.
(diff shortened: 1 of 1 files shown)


RaphaelMeudec:
@iperov Hello, thanks for your PR! However, I don't think this should be a keras-contrib feature. Fix for Adam optimizer should be directy integrated in Keras to avoid a duplicate between keras and keras-contrib. Please consider opening a PR on keras-team/keras !

iperov:
> Adam optimizer should be directy integrated in Keras

I dont think so, because Keras is a standard, that other frameworks should implement. 
Theano, plaidML, and others cannot implement the function of placing and working with tensors on CPU.

Close then.

RaphaelMeudec:
Adam is already on keras [here](https://github.com/keras-team/keras/blob/ad578c4c19444af9d1f0e0d51a8283eb0db1a264/keras/optimizers.py#L433).
Keras-contrib is just an extension of Keras, meant to test new features before eventually integrating them into keras.

PR #483, opened 2019-03-21
Swish activation saves beta as weight even if it is not trainable

### LINKED ISSUE #482: Swish activation doesn't save the weight if beta is not trainable
It is very useful to always save the beta value in the weights file even if beta is not trainable. It is useful when converting to a light weight inference package such as LWTNN (https://github.com/lwtnn/lwtnn).

I already have an implementation of swish activation that preserves all the features of the current one but also saves the untrainable beta in the weights file and I would like to create a pull request.

aghoshpub:
**- What I did**
Allows to save beta as weights when a model is saved even if beta is not trainable.

**- How I did it**
Adding "trainable" argument to the weight.

**- How you can verify it**

---
This pull request fixes #482

### DIFF
--- a/keras_contrib/layers/advanced_activations/swish.py
+++ b/keras_contrib/layers/advanced_activations/swish.py
@@ -1,5 +1,7 @@
-from keras import backend as K
 from keras.layers import Layer
+from keras import backend as K
+from keras.layers import InputSpec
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse:
Closing because of the discussion in #482.

PR #564, opened 2022-10-21
Adding Keras-Contrib functions using string alias

sleepingcat4:
**- What I did**
I have added the method on how to add functions as activations and others from keras-contrib library using string alias

**- How I did it**
I used get_custom_objects() function from keras.utils.generic_utils 

**- How you can verify it**
I added an example snippet code to teach how string alias method can be used to add functions from keras-contrib library in the README.md file. I don't think that requires to write a test as it only shows how to add string alias not creating or compiling an entire model.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,27 @@ model.fit(x=np.random.random((100, 10)), y=np.random.random((100, 100)), epochs=
 # Save our model
 model.save('example.h5')
 ```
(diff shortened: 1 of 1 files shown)


sleepingcat4:
I have updated **advanced functions** importing method in the README file to comply with the latest version of Keras-Contrib

